Issue #116, opened 2015-10-20
Remove invalid values from Finland's name:fin_x_variant


hugovk:
These are all in Finnish, but most of them mean things like "north Finland's", "to Finland", "from Finland", "in Finland", "Finland's" and so on.

Only "Suomen Tasavalta" is needed -- "The Republic of Finland".

```json
name:fin_x_variant: [
"Pohjois-suomen",
"Suomea",
"Suomeen",
"Suomeksi",
"Suomella",
"Suomelle",
"Suomelta",
"Suomen",
"Suomen Tasavalta",
"Suomena",
"Suomessa",
"Suomesta",
"Suometta",
"Suomineen"
],
```

https://whosonfirst.mapzen.com/spelunker/id/85633143/
https://whosonfirst.mapzen.com/data/856/331/43/85633143.geojson

PR #123, opened 2015-10-22
(#116) Remove invalid values and update hierarchy for Finland

hugovk:
* Remove invalid variants from name:fin_x_variant (fixes https://github.com/whosonfirst/whosonfirst-data/issues/116) and name:eng_x_variant
 * Remove a duplicate from name:und_x_variant, it's already in name:fin_x_variant
 * Update hierarchy so the country_id is the country's ID, and not the continent (some other files are like this too)


thisisaaronland:
Thanks!

PR #147, opened 2015-11-04
Correct Astoria Heights spelling

rmarianski:
It looks like "Astoria Hights" was misspelled and should be "Astoria Heights"


nvkelso:
Thank you!

PR #167, opened 2015-11-13
remove local path

missinglink:
fix https://github.com/whosonfirst/whosonfirst-data/issues/166

```bash
$ ls meta/*.csv | xargs sed -i 's|'/usr/local/mapzen/whosonfirst-data/data/'||g'
$ ls meta/*.csv | xargs sed -i 's|'/usr/local/data/whosonfirst-data/data/'||g'
```


missinglink:
is there a unit test suite I can contribute a regression test to?

thisisaaronland:
Thanks - this was a bug where someone (possibly me (possibly in the git pre/post commit hooks)) forgot to pass in flags to enforce relative paths.

I will investigate...

PR #172, opened 2015-11-30
unicode control sequences

missinglink:
some of the data contains strings with unicode control sequences such as:

```bash
M\u00e9tabetchouan\u0096Lac-\u00e0-la-Croix
```

in this case we can simply remove ```\u0096```

this PR is the result of:
```bash
find data -type f -name "*.geojson" | xargs sed -i 's/\\u000[0-9a-f]//g;s/\\u001[a-f]//g;s/\\u00[8-9][0-9a-f]//g'
```

ref: http://www.utf8-chartable.de/
ref: https://en.wikipedia.org/wiki/Unicode_control_characters


thisisaaronland:
Thanks!

PR #191, opened 2016-01-13
Make setup

dphiffer:
This adds a command `make setup` to the Makefile. As of today, the only setup task is to set a configuration that aims to prevent a user of [oh-my-zsh](https://github.com/robbyrussell/oh-my-zsh) from complaining ["oh my so slow!"](https://www.flickr.com/photos/dphiffer/24229559902/). This might also be a placeholder for other one-time setup tasks, should we need them.

To be clear, this doesn't touch the overall performance of `git`, but rather the reporting mechanism built into the Git-integrated PS1 prompt that's turned on by default with oh-my-zsh.


Makefile
```diff
@@ -1,3 +1,13 @@
+setup:
```
nvkelso: Should there be a `all` target that aliases setup? That way someone can type `make` alone on the CL without needing to open up the Makefile to view the target list.

README.md
```diff
@@ -137,6 +137,15 @@ us a measure of authority around the assertions we make about the data
 * Also a way for us to think about how and where we store an audit
 trail (of sorts) for updates to a place
 
+#### Setup script
+
+The repository includes a Makefile at the top level that contains a few file management utilities. Run the following to take care of one-time setup tasks:
```
nvkelso: a few **git** file management utilities?

Makefile
dphiffer: Ah yeah, good point.

README.md
dphiffer: Makes sense, yeah.

PR #229, opened 2016-03-17
Fix Windsor Terrace, Bklyn: 'teraace' to 'terrace'

louh:


thisisaaronland:
Not merging directly because computers are dumb but this changed has been applied by hand. Thanks.

Issue #162, opened 2015-11-12
Capitol Hill neighborhood in Seattle dupes & issues


bcamper:
Numerous issues with Capitol Hill neighborhood in Seattle:

- Has 3 entries:
  - Capitol Hill: https://whosonfirst.mapzen.com/spelunker/id/85809401/
  - Capit**a**l Hill (misspelling): https://whosonfirst.mapzen.com/spelunker/id/85882415/
  - Broadway: https://whosonfirst.mapzen.com/spelunker/id/85807733/

- The one labelled Broadway is the closest representation of Capitol Hill of the 3 (Broadway is the major cultural street running through Capitol Hill, but it is not a neighborhood). 

- The "Capitol Hill" entry is the least accurate of the 3.

- Capitol Hill proper is probably closest to a combination of "Broadway" and "Capital Hill" (sic). Though these still breach multiple surrounding neighborhoods, in particular Montlake to the north.

First priority should be to consolidate and fix spelling.

dphiffer:
Similar kind of situation in Prospect Park South:
https://whosonfirst.mapzen.com/spelunker/id/85865585/
https://whosonfirst.mapzen.com/spelunker/id/85865585/

nvkelso:
Re-opening, because the PR isn't merged yet.

stepps00:
Closing because the PR was merged to master.

PR #261, opened 2016-05-03
Fixed the Seattle neighbourhood of Capitol Hill

stepps00:
Connects to #162.

* Deprecated record 85882415.geojson and record 85809401.geojson with today's date. 
* Added "mz:is_current" field to both records with a value of '0'.
* Updated 85807733.geojson record as true Capitol Hill record. Changed name values from 'Broadway' to 'Capitol Hill' and updated geometry with a combined shape of 85807733.geojson and 85882415.geojson.

**Did not modify zoom level or supersedes/superseded by records.


nvkelso:
I think `id:85807733` should be marked as `placetype=microhood`, parented by the `Capitol Hill` record of `id:85882415`. It's shape should also be merged into that parent record (the shape you propose for 733 should be moved to 415). Don't depreciate 733 or mark it `mz:is_current: 0`, though. 

Additionally, `id:85809401` should be deprecated (e.g. "edtf:deprecated": and "mz:is_current": 0,), as you've proposed. No need to change the shape.

Be careful not to mess with existing properties when they are tied to other datasources so they match up with concordances for GeoNames (e.g. [Broadway](http://www.geonames.org/5788165/broadway.html)), etc, specifically:

```
- "misc:gn_name":"Broadway",
-    "misc:name":"Broadway"
"name:eng_x_preferred":[
 -        "Broadway"
      ],
```

But it's fair game to modify wof properties like:

```
"wof:name":"Capitol Hill",
```

data/858/824/15/85882415.geojson
```diff
@@ -88,7 +90,7 @@
     "wof:parent_id":101730401,
     "wof:placetype":"neighbourhood",
     "wof:superseded_by":[],
-    "wof:supersedes":[],
+    "wof:supersedes":[85809401],
```
nvkelso: @thisisaaronland Is this the correct way to say that 415 gobbled up the geometry for 401 (which was deprecated)?

nvkelso:
Excepting @thisisaaronland approval of https://github.com/whosonfirst/whosonfirst-data/pull/261/files#r61976768, LGTM.

nvkelso:
Per @thisisaaronland, because the place type is changing and that is a significant event, we should make a new record for 85807733 (basically duplicate it) with a new Brooklyn int and set the records WOF id to NEW INT, and supersede 85807733 into that NEW INT.

data/858/824/15/85882415.geojson
nvkelso: @thisisaaronland says "yes" in person :)

data/858/077/33/85807733.geojson
```diff
@@ -83,19 +86,20 @@
...
     "wof:lang":[
         "eng"
     ],
-    "wof:lastmodified":1458857201,
+    "wof:lastmodified":1462321448,
     "wof:name":"Broadway",
-    "wof:parent_id":101730401,
-    "wof:placetype":"neighbourhood",
-    "wof:superseded_by":[],
+    "wof:parent_id":85882415,
```
nvkelso: This should still be `101730401`

data/858/077/33/85807733.geojson
```diff
@@ -83,19 +86,20 @@
...
+    "wof:placetype":"microhood",
```
nvkelso: This should still be `neighbourhood`.

data/858/077/33/85807733.geojson
```diff
@@ -83,19 +86,20 @@
             "country_id":85633793,
             "county_id":102086191,
             "locality_id":101730401,
-            "neighbourhood_id":85807733,
+            "microhood_id":85807733,
```
nvkelso: Remove this.

data/858/077/33/85807733.geojson
```diff
@@ -83,19 +86,20 @@
...
+            "neighbourhood_id":85882415,
```
nvkelso: Set this back to `85807733`

data/858/077/33/85807733.geojson
```diff
@@ -63,11 +65,12 @@
     ],
     "src:lbl:centroid":"yerbashapes",
     "wof:belongsto":[
-        85688623,
+        85882415,
```
nvkelso: Remove

nvkelso:
LGTM, over to @thisisaaronland for final merging.

thisisaaronland:
This will get merged by week's end at the very latest. I would like to see if I can finish getting the Spelunker and all it's associate data inputs migrated to the new servers first. That is really an ideal scenario and if it turns in to quicksand then I will merge-and-deal.

data/772/967/579/772967579.geojson
```diff
@@ -0,0 +1,324 @@
...
+        }
+    ],
+    "wof:id":772967579,
+    "wof:lang":[
+        "eng"
+    ],
+    "wof:lastmodified":1462321448,
+    "wof:name":"Broadway",
+    "wof:parent_id":85882415,
+    "wof:placetype":"microhood",
```
nvkelso: :)

thisisaaronland:
Tell me more about this commit:

https://github.com/whosonfirst/whosonfirst-data/pull/261/commits/4c994767b9008c7de828082a67ce246dbbae571c

We are both deprecating and superseding the record but simultaneously changing its placetype. Shouldn't we just do the first two and not the third ?

thisisaaronland:
@stepps00 - did you see the question above inre: changing the placetype?

stepps00:
@thisisaaronland you are correct. In my [wiki write-up](https://github.com/whosonfirst/whosonfirst-docs/wiki/%23162-Seattle-Neighbourhoods/_edit) about this issue, I outlined the changes to each record - modifying the `wof:placetype` of this feature was not needed for feature `85807733`. I will revert the change in an upcoming commit.

Issue #251, opened 2016-04-18
Clean up Census (balance) artifacts – one feature or many?


nvkelso:
Picking up where #193 leaves off...

```
@martinburch from the Wall Street Journal flagged an issue (pelias/pelias#237) with several 
locality names in WoF/Quattroshapes where the name of the city ends in (balance):

In both OSM and Quattroshapes, some localities end in "balance" as in Athens-Clarke County 
Unified Government (balance)
https://search.mapzen.com/v1/reverse?api_key=xxxxxxxxx&point.lat=33.94806&point.lon=-83.3275&sources=osm

(Example link]

This looks like an artifact of Census-derived data. There are only a handful of these in the 
U.S. Can this be cleaned up somehow?
```

nvkelso:
On the surface, the many parts of a `locality` should just be one feature. But the downside is would be be able to say which **part** of the `locality` is a child of which `county`?

martinburch:
## Indianapolis
I think there is a missing locality entity, `Indianapolis`, which could be a synonym for `Marion` county, simplistically, or could include all Marion's locality children, except the [excluded cities](https://en.wikipedia.org/wiki/Unigov#Excluded_cities_and_town) of 85942579 Beech Grove, 85942567 Lawrence, 85942559 Southport and 85942555 Speedway.

The locality 85942569 Indianapolis city (balance) only exists in Census-land. It could be removed or renamed `Indianapolis`

https://whosonfirst.mapzen.com/spelunker/id/102087201/descendants/?&placetype=locality#10/39.7802/-86.1458

## Athens
For `Athens-Clarke County unified government (balance)`, this shape is synonymous with the city of `Athens`, and that's the most common vernacular name. The hole of 85936181 Winterville is not part of Athens.

https://whosonfirst.mapzen.com/spelunker/id/102081713/descendants/?&placetype=locality#14/33.9587/-83.3689

## Other ` (balance)`
Looking at these two examples, it seems the easiest thing to do would be to rename the ` (balance)` entries to their city names but there might be other case-by-case solutions that more accurately reflect local governance hierarchies.

nvkelso:
For Indianapolis, it's current localadmin is `Center`!?

nvkelso:
[...]

PR #263, opened 2016-05-05
re #251 - Removed (balance) from Augusta-Richmond County

stepps00:


nvkelso:

```
```

nvkelso:
Please don't modify `qs:loc` and `qs:loc_alt` properties, those are original data values from Quattroshapes import that need to be preserved for posterity. 

We should consider using **Augusta** alone for for the `wof:name` (instead of "Augusta-Richmond County"). This is a bit like Washington, D.C. where the city name is "Washington", but the parent geography that is unitary is District of Columbia. Similar to San Francisco where we don't say "City and County of San Francisco" but leave it at "San Francisco".

```
-    "qs:loc":"Augusta-Richmond County consolidated government (balance)",
-    "qs:loc_alt":"Augusta-Richmond County consolidated government (balance)",
+    "qs:loc":"Augusta-Richmond County",
+    "qs:loc_alt":"Augusta-Richmond County",
     "qs:loc_lc":"1304204",
     "qs:pop":0,
     "qs:source":"AUS Census",
@@ -51,7 +51,7 @@
     ],
     "wof:id":85936989,
     "wof:lastmodified":1452296452,
-    "wof:name":"Augusta-Richmond County consolidated government (balance)",
+    "wof:name":"Augusta-Richmond County",
```

nvkelso:
Yep, fixed `qs:loc` and `qs:loc_alt` properties in https://github.com/whosonfirst/whosonfirst-data/pull/263/commits/5241897665ff90ab74d018c1ff06463a2159b237.

nvkelso:
LGTM, over to @thisisaaronland to merge.

PR #265, opened 2016-05-05
re #251 - Removed (balance) from Milford

stepps00:


nvkelso:

```
```

For the `city` component, store that in a new `wof:name` variant for search, ask @thisisaaronland for property name.

nvkelso:
I mean, gotta love auto-generated pages from funky US Census data: 

* http://www.worldatlas.com/na/us/ct/where-is-city-of-milford-balance.html

nvkelso:
LGTM

PR #266, opened 2016-05-05
re #251 - Removed (balance) from Louisville/Jefferson County

stepps00:


nvkelso:

```
```

nvkelso:
Suggest:

`wof:name` to **Louisville**
`wof:label` to **Louisville Metro**

See: 

* https://en.wikipedia.org/wiki/Government_of_Louisville,_Kentucky

stepps00:
Changed:
* `wof:name` to **Louisville**
* `wof:label` to **Louisville Metro**

nvkelso:
LGTM, over to @thisisaaronland to merge.

PR #267, opened 2016-05-05
re #251 - Removed 'city (balance)' from Indianapolis

stepps00:


nvkelso:
```
```

But do modify, `wof:name`.

nvkelso:
Nothing to see here? The `qs` stuff was reverted, but looks like the `wof:name` was already good as **Indianapolis**, so just close this issue?

trescube:
Didn't I fix this with https://github.com/whosonfirst/whosonfirst-data/pull/260?

stepps00:
`qs` field edits were reverted, so no changes were made to this record. Closing the issue because there are no conflicts with the base branch.

PR #279, opened 2016-05-09
Reverted qs changes

stepps00:
#251 - Reverted changes to the qs fields.

Appears as an edit to the qs field, but these values are unchanged from the original. wof:name has (balanced) removed. Kept "Butte-Silver Bow" as full wof:name because sources use full name, instead of just "Butte."


nvkelso:
LGTM. 

In the future, we'd add this commit to an existing Butte balance PR to make it easier to keep track of.

PR #283, opened 2016-05-10
Removed Treasure Island geometry from FiDi

### LINKED ISSUE #126: Does the Financial District (SF) really include Treasure Island
I mean maybe it's like "The City" in London, but probably not...

https://whosonfirst.mapzen.com/spelunker/id/85865899/

screen shot 2015-10-26 at 08 06 46

stepps00:
Fixes #126. Treasure Island geometry was removed from the FiDi record. Requesting merge to Master.


nvkelso:
The new geometry looks good, but...

```
"src:geom":"zetashapes",
"src:geom_alt":[
    "quattroshapes"
],
```

Is now inconsistent. This is a good one for Boundary Issues to think about when supporting geometry uploads, so /cc @dphiffer and @thisisaaronland.

* We should specify `src:geom` as **mapzen** since it's our custom shape now.
* We should store the original zetashape (pre your modifications) into a new alt geom file, and
* We should add **zetashapes** to the `src:geom_alt` list.

stepps00:
Exportified existing geojson file and updated the `src` fields to reflect **mapzen** as the `src:geom` value and **zetashapes** as the `src:geom_alt` value.

Created new zetashapes alt_geometry .geojson and exportified.

data/858/658/99/85865899.geojson
```diff
@@ -50,9 +50,9 @@
     "name:eng_x_variant":[
         "Financialdistrict"
     ],
-    "src:geom":"zetashapes",
+    "src:geom":"mapzen",
     "src:geom_alt":[
-        "quattroshapes"
+        "zetashapes"
```
nvkelso: it should be

```
    "quattroshapes",
    "zetashapes"
```

stepps00: Added in `"quattroshapes"` as a `geom_alt` in latest commit.

nvkelso:
LGTM

thisisaaronland:
👍

PR #299, opened 2016-05-23
show La Lengua earlier, move label south

nvkelso:


thisisaaronland:
I will merge this shortly but this cries out for a "touched by a human" file, so I just want to think about that first before merging and then promptly forgetting about it...

thisisaaronland:
See also: https://github.com/whosonfirst/whosonfirst-data/commit/5d929389096e296d9af5af6e7ef4329785f0f761

cc: @dphiffer

PR #300, opened 2016-05-23
Updated locality records in Mexico City

stepps00:
Fixes https://github.com/whosonfirst/whosonfirst-data/issues/297.

Updated the `wof:placetype` field for the [33 locality records in Mexico City](https://whosonfirst.mapzen.com/spelunker/id/85686515/descendants/?&placetype=locality#12/19.2592/-99.0115). Changed 16 records to `borough`, 17 records to `neighbourhood`.

*    https://en.wikipedia.org/wiki/Mexico_City
*    https://en.wikipedia.org/wiki/Municipalities_of_Mexico_City
*    https://en.wikipedia.org/wiki/List_of_neighborhoods_in_Mexico_City


nvkelso:
Yes, but... supersede the things with new records, then change the `placetype`. Which requires a new tool, @thisisaaronland.

nvkelso:
(Probably kill this PR and make a new one once the new tool lands.)

thisisaaronland:
This is blocked on me writing a proper (or at least "good enough is perfect") supersedes tool. See also:

https://github.com/whosonfirst/whosonfirst-data/issues/301

thisisaaronland:
There are a couple different things going on here...

1. Just out of curiousity, where is Mexico City itself? As in the _city_. Is it https://whosonfirst.mapzen.com/spelunker/id/85686515/ which was incorrectly classified as a region or is it https://en.wikipedia.org/wiki/Greater_Mexico_City and we need to create a record for it or ... ? 

2. In the future when changing placetypes for multiple records it is best to group them by placetype and do the changes for each placetype separately.

@stepps00 Is there a handy list of WOF IDs -> updated placetype list already that doesn't involve me combing through the PR diff?

3. Ultimately yes, per @nvkelso 's comments, this PR will get closed in favour of doing things in small(er) chunks. Related and not-ready-for-prod-yet: 

https://github.com/whosonfirst/py-mapzen-whosonfirst-utils/blob/supersedes/scripts/wof-supersede

stepps00:
@thisisaaronland We do not currently have a record for the _city_ of Mexico City. It was my understanding that the name for this city/region was [changed from **Distrito Federal** to **Ciudad de Mexico**](http://www.theguardian.com/world/2016/jan/29/mexico-city-name-change-federal-district-df) earlier this year and that [Mexico City (the city) and Distrito Federal (the region) were the same entity](https://en.wikipedia.org/wiki/Mexico_City#Federal_District). However, your link to [Greater Mexico City](https://en.wikipedia.org/wiki/Greater_Mexico_City) does make me think there should be a record for the _city_ of Ciudad de Mexico (new record) and a region for Ciudad de Mexico (current record). 

Should that be the workflow? Keep the [current record](https://whosonfirst.mapzen.com/spelunker/id/85686515/_) as a region and create a new record for just the _city_ of Mexico City?

List of WOF IDs with new `wof:placetype` can be found [here](https://github.com/whosonfirst/whosonfirst-data/files/275628/MexicoCity_Localities.txt).

thisisaaronland:
Okay, so my feelings about this are:

* Cuidad de Mexico (WOF ID 85686515) is and has always been a `region` in WOF parlance - that should not change

* "Mexico City" as in the place you visit does not technically exist as an administrative locality the way NYC and its boroughs do; it is effectively a city-state in the classic sense of the term (except for the part where there is still a federal government)

* "According to the 44th article of the Mexican Constitution, in case the powers of the Union move to another city, the Federal District will be transformed into a new state, which will be called "State of the Valley of Mexico", with the new limits set by the Congress of the Union." – Wikipedia

* Anything is possible but assuming Article 44 is ever invoked then the newly created State of the Valley of Mexico would probably need a captial city... like Mexico City. Either way, Mexico City becomes a locality.

* People think of (and expect to find in applications like search) Mexico City as a locality; or put another way it is more like SF or Oakland than it is the "Bay Area" (that would be "Greater Mexico City" which would be more like WOF's yet-to-be-implemented metroarea placetype)

* See also: Tenochtitlan <-- aka Mexico City

I suggest we:

1. Create a new record for a WOF locality called Cuidad de Mexico
2. Parent it by the region of Cuidad de Mexico
3. Assign a `wof:coterminous` property to each record (pointing to the other)
4. Add a `wof:coterminous` entry to https://github.com/whosonfirst/whosonfirst-properties/blob/master/properties/wof.md 
5. Parent the boroughs in the original PR by the new locality

(5) will take some hoop jumping along the lines of everything else in the PR but let's start with 1-4 and then move forward.

Discuss.

stepps00:
Sounds like a valid approach. I will create a new record for the Mexico City locality and update both records with the new `wof:coterminous` field.

thisisaaronland:
@stepps00 - can you make a CSV or plain text file of the cities list:

screen shot 2016-05-25 at 17 37 14

thisisaaronland:
Nothing is being updated yet but dry-runs are chugging along...

```
python ./boroughs.py ./issue-300-borough.csv 
101991115 update 24 children
101991149 update 8 children
101991151 update 29 children
101991185 update 6 children
101991173 update 5 children
101991167 update 21 children
101991181 update 17 children
101991169 update 6 children
101991153 update 23 children
101991175 update 31 children
101991121 update 13 children
101991155 update 1 children
101991157 update 16 children
101991179 update 11 children
101991165 update 3 children
101991171 update 33 children
total 247
```

For details see `whosonfirst-toolbox:scripts/issue-300/boroughs.py`

thisisaaronland:
This is continuing apace but currently blocked on issue #338

thisisaaronland:
@stepps00 these changes should be all rolled in now, independent of the PR. Would you poke around and see if things smell right?

stepps00:
@thisisaaronland looking good! 

So, it looks like the locality records are all superseded correctly and new records with updated placetypes are also correct. I'm noticing that the neighbourhood records in Mexico City have dual hierarchies, though. 

[Nueva Viejo](https://whosonfirst.mapzen.com/spelunker/id/85883665/#16/19.4899/-99.1525), for example, has a dual hierarchy because it is parented by two localities - one to the locality of Mexico City, one to the deprecated locality of Gustavo A. Madero.

Going forward, should we retain dual hierarchies for descendants when deprecating a parent record? My thought would be to remove a hierarchy with a deprecated or superseded record.

stepps00:
Closing because this work was taking care of independently of the PR.

PR #303, opened 2016-05-26
New Ciudad de Mexico locality record

stepps00:
In response to https://github.com/whosonfirst/whosonfirst-data/pull/300. In addition to adding `wof:coterminous` to the [WOF properties page](https://github.com/whosonfirst/whosonfirst-properties/blob/master/properties/wof.md), this PR does the following:

**85686515**
* Updated `wof:lastmodified` field
* Added `wof:coterminous` field with the locality ID as the value.

**857683023**
* Updated geometry for the locality region
* changed `wof:placetype` to `locality`
* Updated ID fields
* Updated hierarchy IDs.


thisisaaronland:
Can you make `wof:coterminous` an array, just to keep things consistent?

stepps00:
Most recent [commit](9aec7a9) changed the `wof:coterminous` field to an array.

PR #314, opened 2016-06-01
Created new SF macrohood records, deprecated the old records

stepps00:
Created macrohood records for the Sunset District, Richmond District, and Downtown in San Francisco. deprecated and superseded old `neighbourhood` records with the same name to the new `macrohood` records. These new `marcohood` IDs are needed as a first step so the parenting can be done for `neighbourhoods` and `microhoods`. Also exportified to clean-up geometries and update `lastmodified`.


burritojustice:
Wondering if the Mission District is a `macrohood`? Thinking it could parent the Mission itself, along with Inner Mission, Mission Dolores, La Lengua and other neighborhoods of interest.

stepps00:
@burritojustice closing this PR without merge, but I'm noting your comment about a Mission `macrohood` in issue #316.

PR #340, opened 2016-06-20
Updated Saint Martin record and dependents

### LINKED ISSUE #13: Promote overseas collectivities of France to dependency (from region of France)
Currently we show this as a region of France, but it is a country of the France empire:

wof_id  iso name
85632737    MF  Saint Martin

data/856/327/37/85632737.geojson

parent now:
85633147
parent proposed:
102191575 (continent), though caveat about France empire

It's also known as: St-Martin

stepps00:
Fixes #13. PR includes the following work:

* Superseded old record of Saint Martin to new record of Saint Martin, vice versa. Grabbed new Brooklyn Int for a new ID.
* New record now has a `placetype` of `country` instead of `region`
* Added `eng_x_variant` name field to new record with value of `St-Martin` (suggested in issue comments)
* Converted strings to integers in two fields of one record for consistency
* Updated `country_id` hierarchy value and `belongsto` value for the two descendants of Saint Martin so no additional superseding work needs to be done related to this issue.
* Updated `lastmodified` fields


data/102/534/449/102534449.geojson
```diff
@@ -40,16 +40,16 @@
             "campus_id":102534449,
             "continent_id":102191581,
             "country_id":85633147,
```
nvkelso: The breaches list is weird, maybe the second element's region should be modified instead of the 1st?

data/102/534/455/102534455.geojson
```diff
@@ -34,12 +34,12 @@
     "wof:hierarchy":[
         {
             "campus_id":102534455,
-            "country_id":85632737,
+            "country_id":874396259,
             "region_id":85673879
```
nvkelso: Does region here get dropped?

data/856/327/37/85632737.geojson
```diff
@@ -110,16 +110,16 @@
         {
             "continent_id":102191575,
             "country_id":85633147,
-            "empire_id":"136253037",
+            "empire_id":136253037,
             "region_id":85632737
```
nvkelso: Does region here get dropped?

stepps00: `region_id` is referencing itself, normally this stays in records.

data/102/534/455/102534455.geojson
stepps00: For this `region_id`, I think this also stays - there is an actual [`region` record of St Martin](https://whosonfirst.mapzen.com/spelunker/id/85673879/#12/18.0778/-63.0788) that is legitimate..

data/856/327/37/85632737.geojson
nvkelso: LGTM

data/102/534/455/102534455.geojson
nvkelso: Ah, that's where I was getting confused :)

data/102/534/449/102534449.geojson
stepps00: Part of the hierarchy - it was split in two (somehow?), so I updated in the most recent commit with correct IDs.

stepps00:
[Most recent commit](e499dc8) updated the `hierarchy` to include the correct IDs. 
All four records in this PR shoulds have correct parenting/hierarchy/`country_id` fields now.

stepps00:
Closing this PR in preference of: https://github.com/whosonfirst-data/whosonfirst-data/pull/458.

PR #365, opened 2016-07-11
Updates to the country of France record

### LINKED ISSUE #364: Update bounding box of France
(original) The [country record](https://whosonfirst.mapzen.com/spelunker/id/85633147) for France contains geometries of France, as well as dependencies (La Reunion, Guadeloupe, etc). The geometry should be updated to only contain France (and Corsica); `lbl` and `geom` values also need to be updated.

(updated) This issue will be used to update bounding box of France country record.

stepps00:
Fixes #364.

*Updated geometry to remove all dependencies from the country record of France. 
* Exportified to update the `bbox`, `lbl`, and `geom` fields.

The [current record](https://whosonfirst.mapzen.com/spelunker/id/85633147) _almost_ looks like what you'd expect hte `empire of France` to look like, but it is still missing Saint Pierre and Miquelon.


nvkelso:
**Semi-related** We're missing one of the constituant parts of the France empire, namely: [French Southern and Antarctic Lands](https://whosonfirst.mapzen.com/spelunker/id/85632549/#3/-32.63/58.66)

nvkelso:
Discussed this in person, the current geom in WOF is "correct" including Alaska and Hawaii equileant overseas regions of France so we don't want to modify the polygon. 

But just like we modified the new `lbl:bbox` for San Francisco and Tokyo, let's add that setting it tight on the continental Europe portion of "France".

Bonus points for setting US label bounding box at the same time tight to the "contiguous 48".

stepps00:
Closing this pull request and opening a new one with the bounding box edits for the US and France. Will note in [the issue](https://github.com/whosonfirst/whosonfirst-data/issues/364).

nvkelso:
@stepps00 please also delete this branch.

PR #370, opened 2016-07-12
Added dependency to French empire record

### LINKED ISSUE #369: Add French Southern and Antarctic Lands to the Empire of France record
Add missing constituent parts of the [France empire](https://whosonfirst.mapzen.com/spelunker/id/136253037/#18/0.00000/0.00000), namely: [French Southern and Antarctic Lands](https://whosonfirst.mapzen.com/spelunker/id/85632549/#3/-32.63/58.66).

stepps00:
Fixes #369.

Added French Southern and Antarctic Lands record to the France empire record.

* Added empire's `wof:id` to the `belongsto` and `hierarchy` field.
* Exportified to update the `lastmodified` date.

Kept `parent_id` as-is, similar to other descendants of empires.


nvkelso:
👍 

I'm not sure how we've worked it for the other members of empires. Do we **only** record that on the country and dependency records, or does it need to be **manually propagated down** to every child in the leaf? For instance: 

- https://whosonfirst.mapzen.com/spelunker/id/85632549/descendants

PR #373, opened 2016-07-20
add known-known about buffered point geometry

missinglink:
add known-known about buffered point geometry


nvkelso:
The proposal for moving this into proper WOF property space is covered in https://github.com/whosonfirst/whosonfirst-data/issues/234.

Issue #383, opened 2016-08-01
Updates in Faroe Islands (cc, geometry)


nvkelso:
1. Looks like we have some bad country codes set in the Faroe Islands.
2. The shape geometry is off by a good bit (from Natural Earth 1:10,000,000 source).

**Some country codes are `FR` but this was changed to `FO` sometime in 80's or or early 90's:**
- https://whosonfirst.mapzen.com/spelunker/id/85633153/descendants/?exclude=nullisland#8/61.890/-7.092

**Shape is off:**

Like other Natural Earth proxy shapes, our shapes for Faroe Islands are significantly off from true location.
- Is there a better open GIS shape source for this?
- If not, move the current shape over onto each island, and adjust major errors by hand.

<img width="1197" alt="screen shot 2016-08-01 at 13 11 32" src="https://cloud.githubusercontent.com/assets/853051/17307386/b19c173a-57e9-11e6-96c6-9d2e2d571978.png">

PR #447, opened 2016-08-15
Faroe Islands updates

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/383. Pull request includes the following updates:
- Updates to the main geometry of the Faroe Islands. This was done by hand, as no other open source shapefile could be found for the islands. The `src:geom` field was updated to `mz` and the existing `naturalearth` geometry was moved to an `alt-geom` file. `naturalearth` was also moved to the `src:altgeom` field.
- Updated _both `wof:country` and `iso:country` fields because the true iso code is `FO`, not `FR` (See: [list of iso codes](https://en.wikipedia.org/wiki/ISO_3166-1).
- Updated `geom` fields, `geomhash` and `lastupdated` fields with exportify tool.


nvkelso:
LGTM.

There are no venues in the descendent list so I don't think any re-PIPing is necessary.

@thisisaaronland this is in response to an external support request thru the Search channel.

thisisaaronland:
If the Faroe Islands are all dependencies of Denmark shouldn't the `wof:country` be "DK" ?

burritojustice:
@thisisaaronland Wouldn't they be part of the Danish Empire, and not a dependency of Denmark?

https://en.wikipedia.org/wiki/Denmark#Greenland_and_the_Faroe_Islands

stepps00:
Hmmm.. looks like the [Faroe Islands record](https://whosonfirst.mapzen.com/spelunker/id/85633153/#8/61.901/-6.960) actually contains a `empire_id` and `belongsto` id for the [Danish Empire](https://whosonfirst.mapzen.com/spelunker/id/136253045/#18/0.00000/-0.00000) already..

thisisaaronland:
`wof:country` means country, not empire.

burritojustice:
I noticed this too -- it seems to point to both country and empire? Shouldn't it just be the latter?
```wof:hierarchy: [
{
continent_id: "102191581",
country_id: "85633121",
dependency_id: "85633153",
empire_id: 136253045
}
],
```

thisisaaronland:
`wof:country` was added to address the issue with the ISO designation that assigns country codes to things that are not countries. It is not for us to tell ISO how to get their party on but it is useful to be able to quickly group/filter records by a standard country code. Hence `wof:country`.

stepps00:
@thisisaaronland - I believe this branch is OK as-is, correct?

We discussed adding the `iso:parent` field with "DK" instead of changing the `wof:country`, but since the Faroe Islands are administered by the [Danish Empire](https://whosonfirst.mapzen.com/spelunker/id/136253045/#18/0.00000/-0.00000) and not [Denmark](https://whosonfirst.mapzen.com/spelunker/id/85633121/#6/56.201/11.613), the `iso:parent` field is not applicable to these records (no iso code for empires).

I can adjust the country codes as necessary, but from my understanding we have all of the correct information in these records.

thisisaaronland:
`wof:country` should be "FR" because the France is the relevant country. `iso:country` should be whatever ISO thinks it is.

stepps00:
Cleaning up, the most recent commit corrected country codes in the Faroe Islands.

No descendants, so we don't have to re-PIP records - should be a clean import.

stepps00:
@nvkelso - OK to merge? this PR should be ready for merge, last commit includes PIP work for all records.

data/101/873/069/101873069.geojson
```diff
@@ -115,14 +119,16 @@
     "wof:geomhash":"2903e7f43099ab269473ec3b29f3bc73",
     "wof:hierarchy":[
         {
-            "continent_id":-1,
+            "continent_id":102191581,
             "country_id":-1,
```
nvkelso: Odd to me we have a `country_id` at all here, since that's mutually exclusive to `dependency_id`. But nothing to hold up this PR over.

nvkelso:
Yep, merge away! :)

Issue #436, opened 2016-08-08
Puerto Rican WOF records have US ISO


trescube:
The following 4 WOF records are in Puerto Rico and have ISO `US`:

https://whosonfirst.mapzen.com/spelunker/id/85633729/descendants/?exclude=nullisland&iso=us#11/18.3851/-66.0980

This differs from the rest of the WOF records (397) in Puerto Rico but have ISO `PR`:

https://whosonfirst.mapzen.com/spelunker/id/85633729/descendants/?exclude=nullisland&iso=pr#9/18.2253/-66.0515

Even though WOF treats Puerto Rico as a dependency and not a country, which is the correct ISO?

thisisaaronland:
Well, the first thing to remember is that `iso:country` != `wof:country`

Specifically `iso:country` should reflect what the ISO peeps think is correct, which doesn't always reflect reality – for example Puerto Rico is not _actually_ a country. `wof:country` reflects the truthiness of countries and the hierarchies.

None of which precludes the data is WOF currently being incorrect.

trescube:
That's fine, it's not a huge deal, I just noticed what appears to be an inconsistency.  I don't think this negatively affects Pelias performance.

stepps00:
Thanks @trescube 

Relatively quick fix, PR is out for this issue. This like the same issue as https://github.com/whosonfirst-data/whosonfirst-data/issues/211.. will revisit soon to crawl WoF records to update these mismatches.

nvkelso:
This PR: https://github.com/whosonfirst-data/whosonfirst-data/issues/453

nvkelso:
Closed via #453.

PR #437, opened 2016-08-08
Updated ISO and WOF country codes from US to PR

stepps00:
Fixes #436.

Updated four records in Puerto Rico with the following:
* `wof:country` changed from `US` to `PR`
* `lastmodified` date updated to reflect date of change

PR #453, opened 2016-08-26
Updates to wof:country code, replaces outdated PR

stepps00:
Replaces PR https://github.com/whosonfirst-data/whosonfirst-data/pull/437

Fixes Issue #436.

Updated four records in Puerto Rico with the following:
- `wof:country` changed from `US` to `PR`
- `lastmodified` date updated to reflect date of change via Exportify


nvkelso:
Changes look good to me. 

Seems to be ready to merge, except meta files aren't updated with the new country code in this PR. @thisisaaronland for that.

thisisaaronland:
Remember `wof:country` is not the same as `iso:country` – and Puerto Rico is a dependency of the US.

I am not opposed to modifying the meaning of `wof:country` to include dependencies but the WOF country label specifically because ISO assigns country code to things that are not.

stepps00:
Only the `wof:country` attributes were changed in this PR (as they were in the original PR) to match the other Puerto Rican features in WoF. The `iso:country` attribute field remained unchanged, as this is an iso-specific value. 

`iso:country` is still `US` and the hierarchy of these four records includes a `dependency_id`. Should we leave the `wof:country` id as-is and add another attribute to signify that these records are part of a dependency?

thisisaaronland:
`wof:country` should be "US" because the United States is the relevant country. `iso:country` should be whatever ISO thinks it is.

stepps00:
@thisisaaronland Puerto Rico has [78 descendant county records](https://whosonfirst.mapzen.com/spelunker/id/85687331/descendants/?exclude=nullisland&placetype=county#9/18.2228/-66.7570). Four of those records (the records in this pull request) have a `wof:country` = `US`, the other 74 records have a `wof:country` = `PR`.

Instead of editing these four records, we should actually be editing the 74 records that have a `wof:country` = `PR`, correct?

Also, I will likely hold off on this issue until https://github.com/whosonfirst-data/whosonfirst-data/pull/484 lands.

nvkelso:
Looks like #484 has been merged, this is ready to action.

stepps00:
Most recent commit (8c12828) fixes merge conflicts.

Issue #391, opened 2016-08-01
Upgrade neighbourhood shapes for Portland, Ore.


nvkelso:
### tl;dr

Status:
- [x] Propose **sources**
- [x] Establish **concordances** between WOF and source
- [x] **Upgrade / add** WOF shapes
- [x] **Add new or missing** features to WOF
- [x] **Deprecate / supersede** WOF features
- [x] **Draft review**
- [x] **Draft revisions** based on review comments
- [x] **Final review**
- [x] **Final revisions** based on review comments
- [ ] **Import** into Who's On First
### Details

Sourcing:
- Permissive license?
- Add new src to Who's On First

Upgrade / add:
- `macrohood`
- `neighbourhood`
- `microhood`

Add new or missing features to WOF
- `macrohood`
- `neighbourhood`
- `microhood`

Deprecate / supersede:
- `macrohood`
- `neighbourhood`
- `microhood`

nvkelso:
**Possible sources:**

- neighborhood associations: 
  - https://www.portlandoregon.gov/oni/article/283289	
  - More detailed: https://www.portlandoregon.gov/bps/article/59267	
- For hoods not on official association-based list: 
  - https://en.wikipedia.org/wiki/Neighborhoods_of_Portland,_Oregon#Other_areas_and_communities	
- Linnton villages/subdivisions: 
  - https://www.portlandoregon.gov/bps/article/116874
- South Portland components: 
  - https://en.wikipedia.org/wiki/South_Portland,_Portland,_Oregon

stepps00:
Consider the updates to the Jefferson Street record, seen [here](https://github.com/whosonfirst-data/whosonfirst-data/pull/454/files) when updating Portland neighbourhoods. Jefferson Street is actually a portion of Downtown; Jefferson Street is not an actual neighbourhood in Portland.

stepps00:
Updated initial comment by @nvkelso after reviewing final Portland neighbourhood geojson files.

Portland neighbourhood geojson files are in the S3 `misc/Issue 391` folder. Files include new macrohood, neighbourhood, and microhood records as well as existing point and polygon features to deprecate.

Over to @thisisaaronland for review and import.

thisisaaronland:
Merged.

PR #454, opened 2016-08-30
Jefferson Street neighbourhood isn't a real thing

ezheidtmann:
This place has a geometry that doesn't overlap the actual Jefferson Street (several blocks south). I have no idea where this name came from but it certainly doesn't describe this area. "Downtown" is a better name but maybe it should be merged with the other Downtown polygon?

Not sure how to contribute this kind of change :-/

---

What problem does this address? - absurd name of neighborhood

What did you see in Who's on First? "Jefferson Street"

What did you expect to see? "Downtown"

Overview of edits and methodology:
Changed name to Downtown throughout.

How was the data updated to address the problem?
Renamed.

How did you perform the edit? (Someone else might want repeat your fix elsewhere, play your fix forward.)
By hand


stepps00:
Hey @ezheidtmann, thanks for the PR and info!

Typically we do update these features through pull requests, but since we have other features (like venues) that rely on the Jefferson Street record for their hierarchy, we need to do more work to update hierarchies, regenerate meta files, etc. Also, this record has been marked as `is_funky: 1`, which means it will not appear in search results or as a map label.

I will likely close this PR, but preserve your notes and add them to [this open issue](https://github.com/whosonfirst-data/whosonfirst-data/issues/391) we have for Portland. We're aware that the [current neighbourhoods of Portland](https://whosonfirst.mapzen.com/spelunker/id/101715829/descendants/?exclude=nullisland&placetype=neighbourhood#11/45.5242/-122.6488) need some work; if you'd like to contribute, feel free to let us know how accurate our potential sources are!

stepps00:
See: https://github.com/whosonfirst-data/whosonfirst-data/issues/391

nvkelso:
Let's land a quick fix for this now (before #391). But let's not merge this PR as-is.

In a new PR (or a modified version of this one), let's create a new Downtown record that is a copy of the old Downtown record (and all it's properties), adding in the additional shape geometry from this weird Jefferson record (almost doubling Downtown's size). 

We'd supersede both old Downtown and Jefferson into the new Downtown, and also mark Jefferson as deprecated since it seems to just be junk.

nvkelso:
@ezheidtmann Thanks much for the tip!

nvkelso:
@ezheidtmann You're correct, the `is_funky` flag only impact map display (not search). We need to deprecate the bad Jefferson record for search to be corrected.

stepps00:
Hey @ezheidtmann, I've opened https://github.com/whosonfirst-data/whosonfirst-data/pull/460 in replace of this PR, which includes changes to additional WoF records that were affected. 

Track the neighbourhood updates in that PR, as this one is now closed.

PR #466, opened 2016-09-13
Updated the three Ramallah records

### LINKED ISSUE #451: merge multiple records for Ramallah PS
https://whosonfirst.mapzen.com/spelunker/id/101921943/

https://whosonfirst.mapzen.com/spelunker/id/101921973/
https://whosonfirst.mapzen.com/spelunker/id/101921953/

Let's keep `101921943` since it has the most number of concordances already.

stepps00:
Fixes #451. 

Included in this PR:
- Added various `x_preferred` names, `megacity`, and `scale` to the Ramallah record that had concordances (as suggested in original issue)
- Superseded / superseded_by work complete to link two invalid Ramallah locality records into the correct locality record
- Deprecated two invalid Ramallah records
- Added `mz:is_current":0` to deprecated records
- Exportified

No descendants, so this should be a straight merge (no required PIP work).


nvkelso:
While I agree that `943` and `973` should be merged together, I don't think `953` should be. 

For `953`, it's across a boundary line and could possibly be a settlement but I don't see any name for the place online. Suggest we deprecate it instead.

nvkelso:
Also, `943` needs to gain the area from `973` (I only see new attributes, not a geometry that is the sum of both records).

nvkelso:
Closing in favor of #532.

Issue #482, opened 2016-09-26
Nashville label appears to be incorrect


trescube:
The record for [Nashville](https://whosonfirst.mapzen.com/spelunker/id/101723183/#10/36.1869/-86.7852), TN has the value for `wof:name` of "Nashville" but it's `wof:label` value is "Nashville-Davidson County".  Is this correct?

stepps00:
Hmm, this is _not_ correct. I thought this issue had to do with an import from https://github.com/whosonfirst-data/whosonfirst-data/issues/217, but that doesn't seem to be the case.

I'll investigate, update the Nashville record, and crawl through other records with `wof:label` fields to ensure they are correct.

stepps00:
Looks like the `wof:label` field was added to the locality record, instead of the county record in [this](https://github.com/whosonfirst-data/whosonfirst-data/pull/264) merged PR.

New PR opened to revert the change (https://github.com/whosonfirst-data/whosonfirst-data/pull/483).

Also, #217 is open and waiting for merge, which will add the `wof:label` field to the actual county record.

PR #264, opened 2016-05-05
re #251 - Removed (balanace) from Nashville-Davidson

stepps00:

PR #483, opened 2016-09-30
Removed label field from Nashville

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/482.

* A prior commit (https://github.com/whosonfirst-data/whosonfirst-data/pull/264) added a `wof:label` field to the locality record of Nashville. This label field was incorrect, this PR reverts that change. The field was removed completely because the `wof:label` field value would be identical to the `wof:name` field value.

* Exportified to update lastmodified date

Should be a quick merge - no other records affected.


nvkelso:
LGTM

nvkelso:
No pipping required.

Issue #430, opened 2016-08-03
Newbury localadmin has geometry, locality doesn't


orangejulius:
Take a look at https://whosonfirst.mapzen.com/spelunker/id/101852663/#13/51.3992/-1.3203

That link goes to the `locality` record for Newbury, UK. That record itself has no geometry (it's a buffered point from Quattroshapes). However, the `localadmin` record for Newbury has what looks like a very nice geometry.

I suspect in this case, the geometries of the two records should actually be identical, so maybe the geometry can just be reused. If so great, if that's not correct, or there's a better solution, that's fine too.

However, this got me thinking about the larger issue. Might there be a lot of places where the `localadmin` has a reasonable geometry, and the `locality` has no geometry? If so, is there a general solution WOF could implement? If not, perhaps Pelias will make the decision to "copy" the geometry from the `localadmin` to `locality` in cases like this. Is that a reasonable idea?

stepps00:
Suggest that we close this issue, as we have a similar issue (#234) open.

The `locality` record for Newbury, UK does have a geometry, but that geometry is in the shape of an ellipse. As suggested in #234, the best option may be to add logic to Pelias to use `localadmin` geometries when a locality's `qs:type = buffered point`.

nvkelso:
If the two records are in fact [coterminous](https://github.com/whosonfirst/whosonfirst-properties/blob/master/properties/wof.md#coterminous), then they should be marked that way in the `wof` properties section. (That's probably the case here but I'd need to research it more, it's fine to link them up now to establish test data.)

If **Pelias** sees that a feature is `coterminous` with another and it's own geometry is approximate, but the coterminous's geometry is not, then I think it'd be okay to use the coterminous feature's geometry. (But not much of that is wired up in the Who's On First data, yet.)

Let's keep this issue open (versus marking it a dup of https://github.com/whosonfirst-data/whosonfirst-data/issues/234), as UK data is more fixable than that in Ireland (which is the general approximate case) because we do have those UK `localadmin` shapes to work from.

@orangejulius Is the UK locality data a huge priority to fix, or is a medium term fix okay? (It's more than a day's worth of work, might be a couple weeks.)

orangejulius:
[...]

Issue #53, opened 2015-08-24
All features should have names


nvkelso:
Bulgaria	85681853
Bulgaria	85681793

stepps00:
Noting that these two records now have name field values (last updated 03.05.2016). 
Are we OK closing this issue?

[Bulgaria 85681853](https://whosonfirst.mapzen.com/spelunker/id/85681853)
[Bulgaria 85681793](https://whosonfirst.mapzen.com/spelunker/id/85681793)

stepps00:
`grep -r '"wof:name":"",' /usr/local/mapzen/whosonfirst-data/data` produced [this list](https://github.com/whosonfirst-data/whosonfirst-data/files/883917/id_w_noname.txt)

Features with ids starting with `109` or `110` are (probably) all from the mesoshapes import and should adopt whatever the `meso:gaul_admin2` name is.

nvkelso:
Eh, 900 features out of 100s of 1000s isn't that bad ;)

stepps00:
Priority of existing properties to use for updating null or blank `wof:name` properties:

1. Use the `wk:page` value in `wof:concordances`:

```
if props.has_key('wof:concordances'):
  concordances = props['wof:concordances']
  if concordances.has_key('wk:page'):
    conc = concordances['wk:page']
    props['wof:name'] = conc
```

2. Use the `meso:gaul_adm2` value:

```
elif props.has_key('meso:gaul_adm2'):
  props['wof:name'] = props['meso:gaul_adm2']
```

3. Use the `name:eng_x_preferred` value:

```
elif props.has_key('name:eng_x_preferred'):
  props['wof:name'] = props['name:eng_x_preferred'][0]

```

4. Use the first `name:eng_x_preferred` value:

```
elif props.has_key('name:eng_x_variant'):
  props['wof:name'] = props['name:eng_x_variant'][0]
  props['name:eng_x_preferred'] = props['name:eng_x_variant'][0]
  del props['name:eng_x_variant'][0]
```

5. Use the `ne:note` value... this should only affect water features in Canada, see [example record](https://whosonfirst.mapzen.com/spelunker/id/404529433/#8/68.337/-106.777).

```
elif props.has_key('ne:note'):
  props['wof:name'] = props['ne:note']
```

6. Else, default to "No data" (outlined in https://github.com/whosonfirst-data/whosonfirst-data/issues/624):

```
else:
  props['wof:name'] = 'No Data'
```

[...]

PR #502, opened 2016-10-20
[wip] Great Britain: Add wof:coterminous fields to buffered point features

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/430.

This PR adds `wof:coterminous` fields to buffered point locality features (and their parent records) with an `iso:country` equal to `GB`.
- A list of locality records with a `"qs:type":"buffered point"` were gathered by running:

```
find /usr/local/mapzen/whosonfirst-data/data -type f -exec grep -l 
'"qs:type":"buffered point",' {} + | xargs grep  '"iso:country":"GB",' > grep.csv
```
- Using this list, a feature collection was created that contained all buffered point features in GB and their parent records. This feature collection projected in QGIS:

screen shot 2016-10-20 at 1 56 39 pm
- If the buffered point locality was parented by something than a localadmin, it was not given a `wof:coterminous` property. If the buffered point locality was parented by a large localadmin record (highlighted in the above image) that had a geometry too large to be coterminous with the locality, it was also not given a `wof:coterminous` property.
- A version of [this script](https://github.com/mapzen/whosonfirst-toolbox/blob/master/scripts/add-field.py) was used to give new `wof:coterminous` properties to the remaining buffered point records and their parent records.


stepps00:
This could be done on a larger scale, possibly for all features with a `"qs:type":"buffered point"`, but logic would be needed to ensure a `wof:coterminous` link wasn't given to a locality record and say, a large region or county record.

nvkelso:
Add specific script variant to Toolbox with new name (in the comments at the top of that script, say it's based on something else...)

nvkelso:
This is a good first cut, but let's be more conservative and _only_ automatically set the `localadmin` and `locality` as coterminous when the names are an exact match.

For instance, [Nettlestone ](https://whosonfirst.mapzen.com/spelunker/id/101852305/) and [Nettlestone And Seaview](https://whosonfirst.mapzen.com/spelunker/id/404445543/) seem like they should **not** be coterminous. The names are slightly different, and imply there are two localities here. 

Note also the poor capitalization in the previous name (`And`), so maybe say exact match on lowercase name (not original string).

nvkelso:
Please also make sure you manually review that the `Newbury` of the primary issue is still resolved after this!

nvkelso:
Related to https://github.com/whosonfirst-data/whosonfirst-data/issues/538.

nvkelso:
Let's defer this one a while yet.

stepps00:
Commit 7cfff413d5d1c2501a2131f4489ac2a9a939fa70 fixes merge conflicts.
Commit e8e63124262a8839d64caa432c39e9ab26527201 converts `wof:coterminous` properties to lists.

stepps00:
Closing this in preference of a new, clean PR. This work is tightly knit with #53.

Issue #155, opened 2015-11-10
Please expose 3-character country codes


trescube:
Pelias uses these codes.  I think this is a duplicate of #103, but I wanted to make it official that search/pelias uses them.

thisisaaronland:
@nvkelso can you think of any reason not to get use 3-letter codes for wof:country?

nvkelso:
As default or as add on? Both should be included somewhere in the data.

stephenkhess:
Both 2- and 3-character ISO codes would be best for Pelias purposes.

thisisaaronland:
Can you walk me through this a bit more? 2 - 3 letter code mappings are readily available and it's not clear what the point of including them both in WOF would be.

stephenkhess:
That's true, there are plenty of sources for the 2->3 letter code mappings for countries, but the idea is that if Pelias is going to start using WOF for administrative data, then it should be *the* source for administrative data.  Pulling in a 3rd party lookup library muddies the water, so to speak.  But we can certainly do that if you think it would be best.

stepps00:
Once the mesoshape imports are complete, I can crawl the whosonfirst-data repo (and venue/postcode repos, if desired) to add in an iso alpha-3 code field to each record. 

This would involve joining metafiles for each of our placetypes to the appropriate iso code in [this csv file](https://github.com/whosonfirst-data/whosonfirst-data/files/520572/iso_codes.csv.zip) and inserting the new iso property field into each record with a script similar to [this one](https://github.com/mapzen/whosonfirst-toolbox/blob/master/scripts/add-field.py). Unless, of course, there is a faster method to add a property field to each record.

stepps00:
Per conversation, the first run of this property addition will be done on all country records. The 3 letter `iso` property should match the existing `iso:country` property.

nvkelso:
(Regarding comment https://github.com/whosonfirst-data/whosonfirst-data/issues/155#issue-116128410, this is not a dup of #103.)

PR #504, opened 2016-10-25
New mz:country property for country records

stepps00:
This PR fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/155.

Using a modified version of [this script](https://github.com/mapzen/whosonfirst-toolbox/blob/master/scripts/add-field.py), a new `mz:country` field was added to all country records in WOF. There was one exception -  [Somaliland](https://whosonfirst.mapzen.com/spelunker/id/85632733/#7/9.773/45.805), which does not have an alpha-3 iso code. PR does the following:
- Adds `mz:country` field based on the country's `iso:country` code
- Updated `lastmodified` date

For now, this field addition will only be done to WOF country records, but country descendants may be updated in the future.

References:
- https://en.wikipedia.org/wiki/ISO_3166-1
- http://www.iso.org/iso/home/standards/country_codes.htm
- http://www.nationsonline.org/oneworld/country_code_list.htm


nvkelso:
@stepps00 Generally yes, but Mapzen doesn't have an opinion about alpha-2 versus alpha-3 codes so I'd prefer this go into `wof:country_alpha3` (paired with existing `wof:country`) instead of `mz:country`. Please update this PR to use that property name instead. Please let Pelias know about the name change.

For Somaliland, Natural Earth has a 3 character code of `SOL`, let's copy that (but keep using `XS` for 2 char default).

nvkelso:
Also, make sure to upload your customized script(s) for this to the Toolbox.

thisisaaronland:
Mapzen _does_ have an opinion about this. It's being done for the Pelias team.

nvkelso:
The proposed 3-char alpha values in `mz:country` _don't_ share the same existing 2-char alpha values as `wof:country`. That's confusing, especially when the existing `iso:country` and `wof:country` _do_ share the same potential range of values. 

I'd rather emphasize the 3-char alpha code lookup / extension of country as `wof:country_alpha3` to establish it's the "same" as `wof:country` but different.

nvkelso:
But it's not different that WOF's opinion, it's just a 3 char code instead of 2 char. Either it should go in wof or Pelias should do the lookup  on their own.

nvkelso:
@stepps00 I talked with @thisisaaronland and we're good to go with `wof:country_alpha3`.

stepps00:
Latest commit removed the `mz:country` property and replaced it with the `wof:country_alpha3` property. Files should be ready for merge to master.

Back to @nvkelso

nvkelso:
LGTM, over to @thisisaaronland to update meta files and merge.

PR #513, opened 2016-11-08
Updated name and official language props for Moscow

### LINKED ISSUE #508: Add wof:lang_x_official and corresponding name to Moscow, RU
In the Moscow, Russia record there is no language specified and the name is in English. Also even if language was specified, and assuming it would be `rus`, there is no corresponding `name:rus_x_preferred` property available. In fact the only correct russian name is under `name:und_x_variant` and I'm not sure what that means.

whosonfirst-data/data/102/003/033/102003033.geojson

stepps00:
Fixes #508.

The locality record for Moscow was missing language properties; this PR fixes that issue and includes the following:

* Adds a new `wof:lang_x_official` property equal to `rus`.
* Adds a new `name:rus_x_preferred` property equal to `\u041c\u043e\u0441\u043a\u0432\u0430` (or `Москва`)
*Updates `lastmodified` date


nvkelso:
LGTM

PR #516, opened 2016-11-10
Mesoshapes Import - Thailand

stepps00:
This PR includes mesoshape features for Thailand and takes care of deprecating existing features.

WOF and mesoshapes had a low join rate based on the `import.py` script logic, so the `meta` files were used to join existing feature IDs to mesoshape features. A variety of name variations were used to join, then double-checked by hand. 

All existing `county` features are point geometries. The `import.py` script logic saved the existing (likely incorrect) `wof:name`s as variant names.


nvkelso:
Spot checked geometries and records – they look good.

Minor name fixes tracked separately in https://github.com/whosonfirst-data/whosonfirst-data/issues/531.

PR #518, opened 2016-11-11
Mesoshapes Import - Russia

stepps00:
This PR includes mesoshape features for Russia and takes care of deprecating existing features.

WOF and mesoshapes had a low join rate based on the import.py script logic, so the meta files were used to join existing feature IDs to mesoshape features. A variety of name variations were used to join, then double-checked by hand. Variations include:

* `sky` to `ski` and `skj`
* Removing ` Raion`
* Removing ` Rayon`
* Joining first ten characters in the `NAME_EN` field (this yielded no extra results)

This resulted in over one hundred duplicate `wof:id` matches, so they were resolved by hand. This brought the deprecated feature rate down from ~92% to ~47%.

All existing county features are point geometries. The import.py script logic saved the existing (likely incorrect) wof:names as variant names.


nvkelso:
Not withstanding some name funk tracked in https://github.com/whosonfirst-data/whosonfirst-data/issues/529, LGTM.

(Missing `region_id` will be added in the PIP step automatically.)

PR #520, opened 2016-11-14
Mesoshapes Import - Japan

stepps00:
PR contains mesoshape features for Japan, as well as deprecated, existing county features.


nvkelso:
We can do a little more name fix-up in https://github.com/whosonfirst-data/whosonfirst-data/issues/527, but the geometries and basic attributes in this issue look good :)

Issue #524, opened 2016-11-15
Create clipped alt-geoms for regions of the US


stepps00:
In continuation of https://github.com/whosonfirst-data/whosonfirst-data/issues/523, Who's On First should have alt-geometries for regions (states) that are clipped to shorelines. 

Two options:

A. Find a new authoritative source that provides these shapes through an open license.

B. Dissolve existing descendant geometries in each affected region into one single shape. Postalcode records are likely the best option because Who's On First already has country-wide postalcode data and the postalcode shapes are clipped at coastlines.

nvkelso:
Related: https://github.com/whosonfirst-data/whosonfirst-data/issues/367

Issue #538, opened 2016-11-29
Towns of Marthas Vineyard: locality versus localadmin


stepps00:
The `localadmin` features on Marthas Vineyard should be reclassified as `locality` features. We are also missing neighbourhood features (ex: Katama). 

https://whosonfirst.mapzen.com/spelunker/id/102084629/descendants/?exclude=nullisland&placetype=localadmin#11/41.3992/-70.8024

nvkelso:
Let's hold off on this until we have a more general plan. One option being considered is to dup some `localadmin` in the USA to have coterminous `locality` features.

nvkelso:
**tl;dr** 

_Proposal 1a:_ Most `localadmin` in 12 New England and upper-midwest United State `region` can be dup'd to form a coterminous `locality`. The remaining features in those 12 states should remain `localadmin` alone, and the `localadmin` features in the other 6 states should remain as-is. This gets pretty complicated internationally – often there are `locality` tagged inside the `localadmin` that aren't coterminous (see anywhere in Europe), no action proposed outside of USA at this time. 

_Propsoal 1b:_ It is still valid for neighbourhoods to be parented by `county` – there are plenty of unincorporated subdivisions or smaller places that we have data for now in WOF that don't have a `locality` parent but _should_ have a `county` parent during PIPing. For the other 6 states, seems like would be useful but not mission critical for `localadmin` to be included in PIPing. It's more of a concern in Europe.

===

The reason New England "town" are excluded from Quattroshapes derived `locality` features (instead they are in `localadmin`) is we got our data from US Census and they specifically prohibit the treatment as New England "towns" as "incorporated place" [[1](https://www.census.gov/geo/reference/gtc/gtc_place.html)] and neither do they treat them as "census designated places". We can of course do something different in Who's On First, but the history is useful to understand how we got here. And this treatment is consistent with what we do in other countries in Europe, Australia, and well the rest of the world.

To recap: Because US Census treats the New England "town" as a "minor civil division", Quattroshapes passed along to us WOF as `localadmin`. 

**Diving deeper...** Who's On First has `localadmin` for 18 regions in the United States:

<img width="928" alt="screen shot 2016-11-29 at 13 39 02" src="https://cloud.githubusercontent.com/assets/853051/20734096/21fce804-b64c-11e6-82d1-c41fdb72b26f.png">

We want to balance what a normal human would consider a `locality` (probably the "town") with also needing to take US Census updates in the future. 

[...]

Issue #367, opened 2016-07-12
Indicate which geometry alt should be used for reverse geocoding


stepps00:
(updated) We should indicate to Point in Polygon (reverse geo) services which of the multiple geometry alts to use as the default geometry is not always the best to use.

In the following example we should use Quattroshapes geometries by defaul, instead of NaturalEarth geometries (this is true for several records). As seen in Canada (below), this caused the descendants of Sumas, WA to be given a hierarchy that contained Canada.

<img width="1175" alt="screen shot 2016-07-11 at 6 13 34 pm" src="https://cloud.githubusercontent.com/assets/18567700/16752094/74a154f0-4793-11e6-8a53-072f1d5c91f3.png">

This is due to the NaturalEarth shapes overlapping international borders (Canada overlapping the US shown below):

<img width="1187" alt="screen shot 2016-07-11 at 6 16 29 pm" src="https://cloud.githubusercontent.com/assets/18567700/16752104/9dd8cea2-4793-11e6-8a38-6b2b6a4b5c37.png">

We should add a new property that specifies which alt-shape to use with PIP service.

nvkelso:
I've updated the title to reflect the data model change. 

@dianashk, @trescube, and @thisisaaronland for PIP reverse geo service implications.

nvkelso:
I propose this new property go in the src property block like so:

`src:rev_geo` = `quattroshapes`

nvkelso:
@thisisaaronland there are a bunch of WOF venues that will need reprocessing, and this faulty PIP results will also affect OSM record decoration in Pelias until it's fixed.

nvkelso:
We'll need to backfill this for `country` place types only at first.

nvkelso:
Here's a white boarding session from last week:

![img_1580](https://cloud.githubusercontent.com/assets/853051/17260566/0866dac0-5586-11e6-9ca0-a3e74442ce1b.JPG)

Even without all that moving around, I think it'd be pretty easy to take the ~10 countries with `ne` sourced geometries and set a new `revgeo` property to `qs` for now to unblock Pelias. It's just a question of what namespace to put it in? Loose in the top-level property or under wof or in a new `geometry` section in properties (unless that's confusing with the geometry object sibling to properties in GeoJSON)?

nvkelso:
Related, generating "display" geometries at 640, 2048, and 4096 resolutions: https://github.com/whosonfirst-data/whosonfirst-data/issues/131

[...]

PR #539, opened 2016-11-30
New US region (state) alt-geometries

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/524

This PR includes new alt-geometry shapes for all regions parented by the United States. These geometries are clipped to coastlines.


nvkelso:
Can you describe where these geometries were sourced and how additional coast clipping was performed, please? When I zoom into the coast it looks low-resolution (like 1:100,000 instead of 1:10,000). See American Samoa or [Pennsylvania](https://github.com/whosonfirst-data/whosonfirst-data/blob/9ca5fb8de5b11c3785d320efe9227f81dc52c024/data/856/884/81/85688481-alt-uscensus.geojson). See image below.

For the filenames, can you add an indication that these geometries are `land` or `clipped` or something? The quattroshape source is actually US Census, too, and includes the land + water, so we need something to differentiate these. See some discussion of this in: https://github.com/whosonfirst-data/whosonfirst-data/issues/367.

screen shot 2016-11-29 at 21 37 06

stepps00:
These alt-geometry shapes were taken directly from the Census' Cartographic Boundaries Shapefiles collection (link [here](https://www.census.gov/geo/maps-data/data/cbf/cbf_state.html)). No additional clipping was performed; the generalized coastline is due to the 500k scale (1:500,000). Three thoughts:

- After tracing the `uscensus` sources file ([here](https://github.com/whosonfirst/whosonfirst-sources/blob/master/sources/uscensus.json)) url link, it looks like that source directs you to a download for the source shapes used by quattroshapes. Since we're using a different Census link and file, are we comfortable appending the new link to the existing source file? Or should a new source be created?

- I propose we use a filename like `85668127-alt-uscensus-500k.geojson`, thoughts?

- If that filename works, I'll need to also include the following in each default record's properties:

``` 
"src:geom_alt":[
        "uscensus-500k"
 ],
````

@nvkelso

nvkelso:
These sound like `display` geometries (see https://github.com/whosonfirst-data/whosonfirst-data/issues/538#issuecomment-263814171) at `zoom-10`, featuring mostly "land" inclusive of terrestrial waters. For Binx they would work, but not for zoomed in map display (we'd want an additional display set at say `zoom-16`, which is the max_zoom we use in Tilezen).

- agree with `uscensus` as _source_
- suggest `display` as the _intended-use_
- suggest `terrestrial` as the _level-of-scope_
- suggest `zoom-10` as the _level-of-detail_
- all together: `85668127-alt-uscensus-display-terrestrial-zoom-10.geojson`
- and in the master record: `"src:geom_alt":[ "uscensus-display-terrestrial-zoom-10"],`

Where geometry clipping (level-of-scope) is one of:

- `original` - from the source, whatever they did
- `territorial` - default optimized for revGeo inclusive of all marine and territorial waters (no water clipping), understood that Mapzen may have modified the original, but it's still "sourced" to the source.
- `terrestrial` - clipped to land + terrestrial waters (e.g. clipped to ocean coastline, but not punched thru for large inland water bodies), understood that Mapzen may have modified the original, but it's still "sourced" to the source. Used for most choropleth maps.
- `land` - no marine waters, no inland waters. Used in contexts where someone doesn't want to load separate "earth" base or "water" top layers. This matches Tilezen's kind value (terrestrial is more the general catch all for the Tilezen layer). The computed label position should be based on this geometry, when available.
- `marine` - geometric opposite of terrestrial, useful for boundary lines.

Here's table converting 1:x representative fractions (eg: `1:500k`) into web map zooms ([src](http://kelsocartography.com/blog/?p=2407)), with some integer rounding:

scales_table

nvkelso:
LGTM

stepps00:
For future reference, script to update the `src:geom_alt` field was imported to the `whosonfirst-toolbox` repo ([here](https://github.com/mapzen/whosonfirst-toolbox/blob/master/scripts/issue524-alt-src-handling.py)). 

The write-up for alt-geometries in there `whosonfirst-cookbook` repo ([here](https://github.com/whosonfirst/whosonfirst-cookbook/blob/master/how_to/creating_alt_geometries.md)) will be updated shortly.

thisisaaronland:
Just to put this in the standard nomenclature for alt geometry names:

* source == `source`
* intended-use == `function`
* level-of-scope and level-of-detail == `extras`, which is an arbitrary array of extras left to the discretion of `source` + `function`

If we need to be precise about what's allowed that information should be encoded in the relevant record in https://github.com/whosonfirst/whosonfirst-sources/

stepps00:
Most recent commit adds the source name (`uscensus`) to the filenames of the alt-geometry files. 

This PR is ready to merge to `master`.

PR #541, opened 2016-12-01
Bounding box updates for FR and US records

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/364#issuecomment-231934449.

As suggested in the issue comments, this PR simply updates the `geom:bbox` fields for United States and France WOF records (Denmark's bbox actually looks okay as-is).


stepps00:
United States:

screen shot 2016-12-01 at 1 52 10 pm

France:

screen shot 2016-12-01 at 1 52 54 pm

data/856/331/47/85633147.geojson
```diff
@@ -6,7 +6,7 @@
     "edtf:inception":"uuuu",
     "geom:area":71.546681,
     "geom:area_square_m":635571946548.695923,
-    "geom:bbox":"-61.797076,-21.370839,55.855019,51.087167",
```
nvkelso: I think the `geom:bbox` is machine generated, and should reflect the true geometric extent. Put this in `lbl:bbox` instead?

data/856/337/93/85633793.geojson
```diff
@@ -5,7 +5,7 @@
     "edtf:cessation":"uuuu",
     "edtf:inception":"uuuu",
     "geom:area":1166.187278,
-    "geom:bbox":"-179.231086,18.86546,179.859681,71.441059",
```
nvkelso: Ditto

data/856/331/47/85633147.geojson
stepps00: They are machine generated; I reverted the change and moved the bbox to the `lbl:bbox` property in most recent commit.

data/856/337/93/85633793.geojson
stepps00: See above ^

nvkelso:
LGTM. I think this can be cleanly merged (property updates only, no hierarchy).

Issue #557, opened 2016-12-13
Merge polygons county Mouscron, Belgium


Platzii:
Hi,

The county of Mouscron is split in 2 parts that do not intersect eachother. However, I noticed these parts have been given a seperate ID.

The polygons of county ID [102049953](https://whosonfirst.mapzen.com/spelunker/id/102049953/#10/50.7385/3.2530) and [102049955](https://whosonfirst.mapzen.com/spelunker/id/102049955/#10/50.7385/3.2530) should be merged into a multipolygon.

Kind regards,
Simon

nvkelso:
@Platzii Thanks for the report, we'll have a look. It'll either be a quick fix for those 2 polygons or something more wide spread to investigate.

stepps00:
Hey @Platzii, thanks for the tip! 

Mouscron is split into two records, though it should actually be one record, as you suggest. This multi-part polygon issue is also affecting Tongeren, records [here](https://whosonfirst.mapzen.com/spelunker/id/102049939/#12/50.7468/5.7965) and [here](https://whosonfirst.mapzen.com/spelunker/id/102049959/#10/50.8645/5.5007). 

With these two duplicate records removed, our `county` counts will match the list of [Arrondissements of Belgium](http://www.statoids.com/ybe.html), so this is likely a quick-fix for Mouscron and Tongeren.

stepps00:
Thanks again, @Platzii - this issue should now be fixed!

PR #560, opened 2016-12-15
Dissolved duplicate county records in Belgium, superseded to new features

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/557. 

Per a user report, this PR merges county records in Belgium. The features were multi-part polygons and were imported to Who's On First as individual records, rather than one combined record. Included in this PR:

* `edtf:cessation` property was updated and given today's date for the existing records. This field was used instead of `edtf:deprecated` because the properties and geometry were correct, but not complete.
*  For the same features, the `mz:is_current` and `wof:superseded_by` properties were updated.
* New records were created with new `wof:id`s and given `wof:supersedes` property values to point back to the incorrect records.
* Existing geometries were merged and used for the new records. They changed from a `polygon` type to a `multipolygon` type.
* Exportified all features to update `geom` and `lastmodified` properties.


nvkelso:
Otherwise LGTM

data/110/878/255/9/1108782559.geojson
```diff
@@ -0,0 +1,77 @@
+{
+  "id": 1108782559,
+  "type": "Feature",
+  "properties": {
+    "edtf:cessation":"uuuu",
+    "edtf:inception":"uuuu",
```
nvkelso: Nevermind: https://github.com/whosonfirst/whosonfirst-properties/blob/master/properties/edtf.md

nvkelso:
Doesn't _need_ to be PIP'd because supersedes pointers are there, but eventually it should be re-PIP'd.

Issue #70, opened 2015-08-31
admin-2-counties have ALL CAPS names


nvkelso:
Some geographies have place names in all caps.

For instance SERRA TALHADA (wof_id: 102056839) in Brazil should be something more like Serra Talhada.

Reported by @missinglink

missinglink:
duplicate: https://github.com/whosonfirst/whosonfirst-data/issues/164

stepps00:
Suggest we close this issue in preference of #164, which will update _all_ records, including any records with an admin2-level placetype.

PR #569, opened 2016-12-27
Mesoshapes county 1.0 cleanup - Indonesia

stepps00:
This PR cleans up the mesoshapes import in Indonesia. Specifically, this PR:

* Deprecated `county` features that were not deprecated in the original mesoshapes import PR for Indonesia.
* Imports "leftover" `county` features that were not included in the original mesoshapes import PR for Indonesia. Features were manually mapped to existing `wof:id`s in the original import PR, which caused the import script to skip 97 features when importing. This was fixed in commit 5e5db53.


data/110/879/677/9/1108796779.geojson
```diff
@@ -0,0 +1,50 @@
...
+    "wof:geomhash":"7ee81cbae5012231e441e1b174a2fe77",
+    "wof:hierarchy":[
+        {
+            "country_id":85632203,
+            "county_id":1108796779
+        }
+    ],
+    "wof:id":1108796779,
+    "wof:lastmodified":1482878476,
+    "wof:name":"SUBULUSSALAM",
```
nvkelso: All the `"wof:name"` properties should be Title Cased. Either fix in this PR or create new issue to track fix post-import.

nvkelso: (In this case, `"meso:gadm_name":"Aceh Singkil"` and `"meso:gaul_adm2":"Aceh Singkil"` have the correct casing, but other records have DIFFERENT names in those properties, so we'll need to change the case of `"wof:name"` instead of sourcing it from a different field.)

nvkelso: Already tracked in: https://github.com/whosonfirst-data/whosonfirst-data/issues/70

thisisaaronland:
Processing in dry-run mode...

thisisaaronland:
Processing for realz...

Issue #570, opened 2016-12-29
Hierarchy issues


dianashk:
During our regularly-scheduled Pelias builds we ran into several regressions that seem to be related to WOF admin hierarchies. Here is what we found.

1. When searching for `New York, New York` we were expecting it to be parented by Richmond County but are now seeing Kings County   [see here](http://pelias.github.io/compare/#/v1/search%3Fsources=wof&text=New%20York,%20New%20York) and [85977539](https://whosonfirst.mapzen.com/spelunker/id/85977539/#9/40.6979/-73.9797)

2. Another NY county update when searching for Manhattan, used to be within New York County and is now in Kings County [see here](http://pelias.github.io/compare/#/v1/search/structured%3Fborough=Manhattan&locality=New%20York&region=NY) and [421205771](https://whosonfirst.mapzen.com/spelunker/id/421205771/#11/40.7809/-73.9772)

3. This county in China has been deprecated. Any particular reason? Was it superceded by something? [890513061](https://whosonfirst.mapzen.com/spelunker/id/890513061/#6/30.215/107.730)

4. These two records have lost parentage: [85763987](https://whosonfirst.mapzen.com/spelunker/id/85763987) and [85793671](https://whosonfirst.mapzen.com/spelunker/id/85793671)

nvkelso:
We'll look at these today and expect to propose fixes by EOD.

nvkelso:
@dianashk 

- Pelias should update their test for `wof:id` = `890513061` to use `wof:id` = `890513063` instead. See discussion in the PR.
- All the other changes will be addressed in the PR.

Issue #238, opened 2016-03-29
"NYC" and metro areas


thisisaaronland:
Attention @nvkelso @riordan @trescube @dphiffer 

The short version is: The locality of New York City (85977539) is parented by the local admin or borough of Brooklyn (404521211).

The problem is: This is neither correct nor incorrect. There are a series of overlapping issues:

1. The polygon for NYC which is broadly correct for the 5 boroughs isn't really a "locality"
2. The label centroid for that polygon, as derived by mapshaper is in fact in Brooklyn so the PIP code is correct in saying the centroid is in Brooklyn
3. https://whosonfirst.mapzen.com/spelunker/id/404522731/ <— the local admin / borough of Manhattan
4. The part where NYC should really have multiple hierarchies – namely all 5 boroughs – or:
5. The part where NYC should probably not be classified as a locality at all, but rather a metro area
6. The part where people refer to “NYC” as only manhattan or all five boroughs depending on your view of the world
7. The part where people refer to cities and metroareas more or less interchangeably and often with diametrically opposite meanings depending on whether they are a local or a tourist

My inclination would be to finally pull the trigger on "metroareas" and:

1. Supersede 85977539 (NYC, the locality) with a new record for NYC, the metroarea
2. Parent the 5 boroughs (localadmins) with the new record
3. Merge most or all of the exonyms for NYC and Manhattan (there aren't any yet)
4. Update all the things to include the metro area - this is a bit daunting but it won't be the last time we'll need to do this so...

Immediate concerns:

* Will this make the Pelias build sad because it doesn't know about “metroareas” ?

Discuss.

dphiffer:
In no particular order:
* Manhattan equated with NYC feels so wrong, although I understand that wrongness may strike visitors as misplaced (i.e., when the tourist types in "NYC" they may not actually *mean* all 5 boroughs). Even so, I would be really sad if this is how the data described the city.
* I might actually prefer 5 hierarchies to a metroarea, if only because elsewhere "New York Metropolitan Area" tends to include other nearby cities. It's messier though.
* Basically I like how Ingrid describes it in her Twitter bio: "A Small Island Near America"

dphiffer:
Or maybe it's "a set of small islands." Did I just equate Manhattan with NYC?

nvkelso:
[...]

PR #571, opened 2016-12-29
Mesoshape and mystery Pelias fixes

nvkelso:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/570.


data/421/205/771/421205771.geojson
```diff
@@ -30,14 +30,16 @@
...
+    "reversegeo:longitude":-73.968777,
     "src:geom":"quattroshapes",
     "wof:belongsto":[
         85688543,
         102191575,
         404521211,
         85633793,
         85977539,
-        102082361
+        102081863
```
nvkelso: Change LGTM. 

Background: we think this error originates because the parent is the city of NY, which is parented by Kings (because the locality's math latlng falls in Kings)

data/859/775/39/85977539.geojson
```diff
@@ -730,7 +732,7 @@
         {
             "continent_id":102191575,
             "country_id":85633793,
-            "county_id":102082361,
+            "county_id":102081779,
```
nvkelso: We should have the list of 5 county hierarchies from https://raw.githubusercontent.com/whosonfirst-data/whosonfirst-data/78c40d0e8ce5a21271b77254d5ecc6ac5bf6f6e1/data/859/775/39/85977539.geojson.

nvkelso: @thisisaaronland Looks like the re-PIPing destroyed the earlier manual fix to add the multiple hierarchies in https://github.com/whosonfirst-data/whosonfirst-data/issues/238 :(

data/859/775/39/85977539.geojson
```diff
@@ -708,7 +710,7 @@
         102191575,
         404521211,
         85633793,
-        102082361
+        102081779
```
nvkelso: Similarly, this would need to include all 5 counties, not just 1.

data/859/775/39/85977539.geojson
nvkelso: @dianashk We think your test has been passing since March 29th using the multiple hierarchies when our #238 landed, but there is a chance you may need to update your test.

data/857/639/87/85763987.geojson
```diff
@@ -72,14 +78,21 @@
     },
     "wof:country":"PA",
     "wof:geomhash":"3523da25fdcb70ee20a09d0811fe7286",
-    "wof:hierarchy":[],
+    "wof:hierarchy":[
+        {
+            "continent_id":102191575,
+            "country_id":85632179,
+            "neighbourhood_id":85763987,
```
nvkelso: All the new hierarchy looks good!

I'm surprised the `neighbourhood_id` wasn't included in the hierarchy at all, even if a parent wasn't found.

@thisisaaronland Looks like our initial commit for this feature used a different PIP workflow where if the parent placetype wasn't found, it would look for the next, and continue up to continent?

thisisaaronland: This is the "ensure_hierarchy" code that was added. Previously, if a parent ID could not be identified we gave up trying to populate the hierarchy. If the `ensure_hierarchy` flag is passed then even if parent_id is `-1` we walk up the stack trying to find the first containing ancestor and build the hierarchy from there.

nvkelso:
@stepps00 

1. NYC needs changes listed above.
2. What about the county in China, `890513061`?

nvkelso:
For China, we did some digging: 

- Pelias should update their test to use `wof:id` = [`890513063`](https://whosonfirst.mapzen.com/spelunker/id/890513063/#6/30.215/107.730).
- Why: It's valid to have deprecated this feature (no change in this PR). 
- Background: Looks like WOF imported two point geometry records from Quattroshapes, which got those 2 records from GeoPlanet, and Flickr which had duplicate places & different photos. Those two point geoms had the same exact names and hierarchy, but different photo counts. When we ran the mesoshape import concordance script, it found one of them, and marked the other as no-match for deprecation. No attempt was made to untangle records or do any superseding work (as in most cases it should only be deprecated). 
- We _could_ have added a `superseded_by` to `890513061` (pointing at `890513063`, and vis-a-versa), but we think the Pelias test would still have failed, and we don't recommend adding this superseding relationship now. (In the future for one-off edits we probably will, just not for massive mesoshape imports).

data/859/775/39/85977539.geojson
thisisaaronland: Unsure what happened here but it appears to be correct here:

```
python ./test-hier.py -D /usr/local/data -R whosonfirst-data -I  85977539
BEFORE 404521211 (parent ID)
[{u'continent_id': 102191575,
  u'country_id': 85633793,
  u'county_id': 102082361,
  u'localadmin_id': 404521211,
  u'locality_id': 85977539,
  u'region_id': 85688543}]
INFO:root:append parent and hierarchy for 85977539
INFO:root:find parent (localadmin) for 40.664912, -73.895946 : 0
INFO:root:find parent (county) for 40.664912, -73.895946 : 1
INFO:root:parent ID has changed for 85977539
AFTER 102082361 (parent ID)
[{u'continent_id': 102191575,
  u'country_id': 85633793,
  u'county_id': 102082361,
  u'locality_id': 85977539,
  u'region_id': 85688543}]
```
This is just a tool for testing what _would_ happen. No changes have been applied, yet.

stepps00:
@nvkelso commit  	2ec9f93 updates the remaining NYC hierarchy/belongsto issue. All four failed Pelias should now be taken care of in this PR.

@thisisaaronland when a dual hierarchy is hand-built in a record (like [the one seen here]( 	https://github.com/whosonfirst-data/whosonfirst-data/commit/2ec9f93abc1a7d592315cc363b6394f913bdecd6)), is there a property flag to pass during the PIP process to make sure the hierarchy is not reset?

thisisaaronland:
@stepps00 - add `wof:hierarchy` to the `wof:controlled` property for that record. Details are here: https://github.com/mapzen/whosonfirst-toolbox/blob/master/scripts/pip-mesoshapes/hierarchy.py#L227-L241

data/857/936/71/85793671.geojson
```diff
@@ -62,14 +68,21 @@
     },
     "wof:country":"CU",
     "wof:geomhash":"99778fe3f227118a94d00d783b5c24fa",
-    "wof:hierarchy":[],
+    "wof:hierarchy":[
+        {
+            "continent_id":102191575,
+            "country_id":85632675,
+            "neighbourhood_id":85793671,
+            "region_id":85670441
```
stepps00: @nvkelso New issue filed here: https://github.com/whosonfirst-data/whosonfirst-data/issues/572

Issue #599, opened 2017-01-09
Character encoding issue with WrocÅawski - Wrocławski in Poland


rmglennon:
https://whosonfirst.mapzen.com/spelunker/id/102079685/#9/51.0436/16.9712 

A user reported "county" "WrocÅawski" (with an A) should be "Wrocławski" (in English, it's an L with a stroke)

Issue #183, opened 2016-01-05
backfill `wof:name` to always be 7-bit ASCII


thisisaaronland:
Update export/validation tools accordingly.

thisisaaronland:
For dry-run testing if nothing else:

* https://github.com/rainycape/unidecode
* https://pypi.python.org/pypi/Unidecode/

thisisaaronland:
The (Go) unidecode package seems to do okay, until it doesn't...

```
85667819,Gjirokastër,Gjirokaster
85667821,Dibër,Diber
85667829,Lezhë,Lezhe
85667831,Durrës,Durres
85667835,Finström,Finstrom
85667849,Eckerö,Eckero
85667857,Föglö,Foglo
85667783,Durrës,Durres
85667793,Shkodër,Shkoder
85667797,Kukës,Kukes
85667945,Sant Julià de Lòria,Sant Julia De Loria
85667885,Vårdö,Vardo
85667899,Brändö,Brando
85668011,Neuquén,Neuquen
85668025,Río Negro,Rio Negro
85668031,Córdoba,Cordoba
85668059,Tucumán,Tucuman
85668073,Entre Ríos,Entre Rios
85668209,Gəncə,G@Nc@
85668221,Kəlbəcər,K@Lb@C@R
85668225,Mingəçevir,Ming@Cevir
85668195,Ağstafa,Agstafa
85668199,Daşkəsən,Dask@S@N
85668313,Beyləqan,Beyl@Qan
85668317,Biləsuvar,Bil@Suvar
85668323,Cəbrayıl,C@Brayil
85668331,Cəlilabad,C@Lilabad
85668339,Dəvəçi,D@V@Ci
```

No idea about CJK:

```
102032007,Misaki Chō,Misaki Cho
102032009,Tōya,Toya
102032027,Ōmu,Omu
102032029,Ōdate,Odate
85916145,十ノ通り,Shi NoTong Ri
102032049,Kamikitakōzan,Kamikitakozan
102031661,Kami-gōchi,Kami-Gochi
102032079,Kōya,Koya
85916149,十二ノ通り,Shi Er NoTong Ri
85916153,四の通り,Si NoTong Ri
102032101,Bandō,Bando
85916157,３丁目,3Ding Mu 
102032111,Rittō,Ritto
102032115,Kōka,Koka
102032139,Ōnojō,Onojo
102032141,Minami-Sōma,Minami-Soma
[...]

PR #600, opened 2017-01-09
Updated wof:name and name property for county in Poland

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/599.

- Updated the `wof:name` to read "Wrocławski", rather than "WrocÅawski".
- Moved "WrocÅawski" to the `"name:pol_x_preferred"` property.
- Exportified

Sources:
- https://en.wikipedia.org/wiki/Wroc%C5%82aw_County
- https://international.uni.wroc.pl/en
- http://www.statoids.com/ypl.html


data/102/079/685/102079685.geojson
```diff
@@ -14,6 +14,9 @@
     "mps:latitude":50.971294,
     "mps:longitude":16.840011,
     "mz:hierarchy_label":1,
+    "name:pol_x_preferred":[
+        "Wroc\u00c5awski"
```
nvkelso: This should be `Wrocławski` (with the slash thingy in the "l" character).

data/102/079/685/102079685.geojson
```diff
@@ -47,8 +50,8 @@
         }
     ],
     "wof:id":102079685,
-    "wof:lastmodified":1466628619,
-    "wof:name":"Wroc\u00c5awski",
+    "wof:lastmodified":1483997979,
+    "wof:name":"Wroc\u0142awski",
```
nvkelso: This one should be ASCII 7, so just a basic `l` character.

data/102/079/685/102079685.geojson
nvkelso: (eg: `"Wroc\u0142awski"`)

data/102/079/685/102079685.geojson
nvkelso: This won't completely satisfy the original issue, but is correct from WOF's perspective. Since Pelias doesn't support looking in the alternative names, we could keep what you've proposed for the default `"wof:name"`, and scrub it later as part of https://github.com/whosonfirst-data/whosonfirst-data/issues/183.

data/102/079/685/102079685.geojson
stepps00: I'll update the `preferred` value to be `"Wroc\u0142awski"`, but we still need to retain the original name (`"Wroc\u00c5awski"`); move this to the `pol_x_variant` field?

data/102/079/685/102079685.geojson
stepps00: Taken care of in most recent commit.

data/102/079/685/102079685.geojson
stepps00: Taken care of in most recent commit (kept original name in `variant` field).

data/102/079/685/102079685.geojson
```diff
@@ -14,6 +14,12 @@
     "mps:latitude":50.971294,
     "mps:longitude":16.840011,
     "mz:hierarchy_label":1,
+    "name:pol_x_preferred":[
+        "Wroc\u0142awski"
+    ],
+    "name:pol_x_variant":[
+        "Wroc\u00c5awski"
```
nvkelso: Let's not preserve encoding trash as a variant name. (In this case it wasn't a common misspelling but character encoding funk that should be purged.)

nvkelso: (Let's ignore strict life cycle rules for this, keep in this one record – no new ID.)

stepps00: Removed in:  	83bf3e7

Issue #575, opened 2017-01-03
Duplicate localities in Bratislava, SK 


rmglennon:
It looks like there may be a duplicated locality of Berg in Bratislava, Slovenia. It appears the polygon is split (possibly across Bratislava district boundaries, known as a "county" in WoF) and needs to be merged into one. 

Berg may or may not be in the proper admin hierarchy, though. Locality filter for Bratislava shows the duplicated Berg entries:

https://whosonfirst.mapzen.com/spelunker/search/?q=bratislava&placetype=locality#12/48.1511/17.1666

This started from a user support request. Snips: "you are using very often in Bratislava, the capital of Slovakia, town/translation (not sure) Berg instead of Bratislava, it may have something to do with one of your source.

Sample latLngs: lat: 48.15984513772255; lng: 17.109060287475586"

stepps00:
Scrub: https://whosonfirst.mapzen.com/spelunker/id/85688373/descendants/?exclude=nullisland&placetype=locality#9/48.3152/17.3131

nvkelso:
This smells like NYC which is a locality with multiple hierarchies, one for
each county it crosses?

pavzi:
None of the polygons with name Berg in this link https://whosonfirst.mapzen.com/spelunker/id/85688373/descendants/?exclude=nullisland&placetype=locality#9/48.3152/17.3131 are Bratislava. 

Berg is small village in Austria near to the border with Slovakia: https://www.google.sk/maps/place/Berg,+Rakousko/@48.1042523,17.0140854,14z/data=!3m1!4b1!4m5!3m4!1s0x476c8a6ce5fcb3fb:0xd85cfbc9ad1f8537!8m2!3d48.1033368!4d17.0409912

stepps00:
Thanks @pavzi - we'll investigate and update our records accordingly.

stepps00:
After researching this area and comparing our records, it looks like the five "Berg" records are all part of "Bratislava". A screenshot of all Berg geometries and the Bratislava geometry projected in QGIS shown below:

<img width="1220" alt="screen shot 2017-01-10 at 4 12 47 pm" src="https://cloud.githubusercontent.com/assets/18567700/21830216/0e60865a-d750-11e6-938c-b69c5b652a0e.png">

[...]

PR #609, opened 2017-01-11
Merged Berg and Bratislava records in SK

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/575.

- A new record with the merged geometries of all Berg and Bratislava records.
- `superseded_by` and `supersedes` linking
- `"mz:is_current"=0` property additions for all records
- Cessation dates for all records; used this property instead of deprecated because the geometries/records were all _correct_, just not classified correctly
- Exportified


nvkelso:
LGTM

Issue #618, opened 2017-01-16
Merge Jämtland multi-polygon region in Sweden


nvkelso:
Sweden should only have 21 top-level `region` features, but we report 22 due to a duplicate `Jämtland`.

- The good one: https://whosonfirst.mapzen.com/spelunker/id/85688461/#6/63.397/14.487
- The leftover sliver that should be merged with `461`: https://whosonfirst.mapzen.com/spelunker/id/85688443/#13/62.0201/15.3904

nvkelso:
Related: 

- https://github.com/whosonfirst-data/whosonfirst-data/issues/617

stepps00:
The current [Jämtland record](https://whosonfirst.mapzen.com/spelunker/id/85688461/#6/63.397/14.487) now contains the geometry of the smaller, deprecated [Jämtland record](https://whosonfirst.mapzen.com/spelunker/id/85688443/#13/62.0201/15.3904).

The superseding work/deprecating happened in a prior commit (https://github.com/whosonfirst-data/whosonfirst-data/commit/5941e41a0a98dc9643d71c464db08d0b61312422). Closing issue.

stepps00:
Noting that Sweden also correctly has 21 `region` descendants now, too.. 

https://whosonfirst.mapzen.com/spelunker/id/85633789/descendants/?exclude=nullisland&placetype=region

PR #619, opened 2017-01-17
Merged two Sweden region records of Jämtland

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/618.

- Merges geometries of two Jämtland `region` records in Sweden into one record (`461`)
- Deprecates, adds `superseded_by` value, and adds `is_current` field to smaller geometry (`443`)
- Adds `supersedes` value to correct record (`461`)
- Ran updated record (`461`) through Mapshaper for new `mps:lat/long` and `lbl:lat/long` properties
- Exportified


data/856/884/61/85688461.geojson
```diff
@@ -76,12 +76,14 @@
...
     ],
     "wof:id":85688461,
-    "wof:lastmodified":1467242952,
+    "wof:lastmodified":1484673730,
     "wof:name":"J\u00e4mtland",
     "wof:parent_id":85633789,
     "wof:placetype":"region",
     "wof:repo":"whosonfirst-data",
-    "wof:superseded_by":[],
+    "wof:superseded_by":[
```
nvkelso: Should this `id` be supersedes instead?

nvkelso:
LGTM

nvkelso:
(Also double checked and neighboring `region` does correctly have a [hole](https://whosonfirst.mapzen.com/spelunker/id/85688453/#9/61.9028/15.5237) for that exclave.)

Issue #208, opened 2016-02-17
New French 2016 regions


nvkelso:
From the Geonames list:

```
From 1st January [2016-ed] France has a new division
(https://en.wikipedia.org/wiki/Regions_of_France).
```

stepps00:
The **WOF S3 Dropbox misc folder** now contains a file with new geometries for 2016 French regions, [link here](https://s3.amazonaws.com/whosonfirst.mapzen.com/misc/Issue%20208/Import_NewRegions.geojson). Contents/attributes:

The geojson file contains seven new geometries that should be imported as new regions in France; these shapes were created by combining existing French regions. Since the restructuring of French regions [did not change any borders](https://en.wikipedia.org/wiki/Regions_of_France#Reform_and_mergers_of_regions), this should be a simple supersedes/supersedes_by task.

**wof:name**
* New `wof:name` field added for each new region. Sources [here](https://en.wikipedia.org/wiki/Regions_of_France) and [here](http://www.insee.fr/fr/methodes/nomenclatures/cog/).

**name:fra_x_preferred**
* If the new region had a non-English transliterated name (_ex: Île-de-France_), it was added to this field as a new `name` field.

**wof:supersedes**
* This string field contains any `wof:id`s of any records that the new record supersedes.

nvkelso:
LGTM

stepps00:
Files removed from S3; the PR mentioned above should include all work required to fix this issue.

PR #655, opened 2017-02-16
Updated macroregion records in France

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/208.

This PR updates the geometries and records for `macroregion` records in France. On 01-01-2016, the `macroregion` records in France were consolidated. This PR includes the following work:

- Merged geometries of pre-2016 macroregions to create geometries for new macroregion records. Imported these merged geometries in new macroregion records.
- Added `wof:supersedes` property values to new records 
- Added `wof:superseded_by`, `mz:is_current`, and `mz:cessation` property values to outdated records

Files removed from S3, this PR should include all required updates.


nvkelso:
Will require PIPing.

stepps00:
@thisisaaronland - I'm not seeing these 23 new French `macroregion` records in the Spelunker, but I do see them in GitHub ([here](https://github.com/whosonfirst-data/whosonfirst-data/blob/master/data/110/882/639/3/1108826393.geojson)) and in S3. Any idea why that is?

thisisaaronland:
Maybe just bunk indexing. I can re-run that by hand.

stepps00:
Note: All records updated in this PR now appear in the Spelunker.

PR #657, opened 2017-02-17
Spain: Canary Islands record updates

### LINKED ISSUE #201: Canary Islands are missing all hierarchy
https://whosonfirst.mapzen.com/spelunker/id/404227369/

stepps00:
Fixes #201.

Updated Canary Islands record with `reversegeo:*` lat/long properties and ran against PIP tools to update descendant records.


data/404/227/369/404227369.geojson
```diff
@@ -43,14 +48,15 @@
     "wof:geomhash":"a6616ff66b0c08344fa3d1e4afeb43d7",
     "wof:hierarchy":[
         {
+            "continent_id":102191573,
```
nvkelso: I think this needs to be controlled property.

stepps00: Updated record with a `controlled` prop in commit 734c581.

Over to @thisisaaronland.

Issue #673, opened 2017-02-25
Update geometries for Airport campus records


stepps00:
In previous areas where we have completed neighbourhood upgrade work (and for all future neighbourhood imports), the geometry of each campus representing an airport should have a new geometry, which should be the same shape as the neighbourhood feature for that airport district. 

As an example, the campus record for the [Portland International Airport](https://whosonfirst.mapzen.com/spelunker/id/102532251/#13/45.5848/-122.5826) should actually have a geometry equal to the [Portland International Airport](https://whosonfirst.mapzen.com/spelunker/id/1108721307/#13/45.5840/-122.5910) neighbourhood.

Go back through each market, update the Airport campus record in a branch, and propose a PR.

nvkelso:
For new workflow we’ll want to add a (proposed) geom_id`  property and set
it to the airports campus `wof:id` value, and @thisisaaronland will need to
add new feature to import script to support that.

nvkelso:
For new workflow we’ll want to add a (proposed) `geom_id` property and set
it to the airports campus `wof:id` value, and @thisisaaronland will need to
add new feature to import script to support that.

thisisaaronland:
I am going to parrot back what I think I understand here:

* If I see a property called `geom_id` (and I would like to propose that it be `mz:geom_id` at least by the time it reaches code (so something something something property aliases) just to ensure that everything is namespaced) then:

* Update the geometry of the thing referenced by `mz:geom_id` with the current geometry
* Remove the `mz:geom_id` property (before exporting the current record)

Correct?

nvkelso:
Yes :)

stepps00:
FYI - [properties alias](https://github.com/whosonfirst/whosonfirst-properties/blob/master/aliases/property_aliases.json#L212) and the [external editor option](https://github.com/whosonfirst/py-mapzen-whosonfirst-utils/blob/master/scripts/wof-csv-to-feature-collection#L104) have been updated to include the `mz:geom_id` prop.

Issue #672, opened 2017-02-25
Append "District" to the end of Airport neighbourhood names


stepps00:
In previous areas where we have completed neighbourhood upgrade work (and for all future neighbourhood imports), the `wof:name` of each neighbourhood representing an airport should have ` District` appended to the end of the name. 

As an example, the record for the [Portland International Airport](https://whosonfirst.mapzen.com/spelunker/id/1108721307/#13/45.5840/-122.5910) neighbourhood should actually have a `wof:name` of `Portland International Airport District`.

Go back through each market, update the Airport neighbourhood record in a branch, and propose a PR.

nvkelso:
@stepps00 for markets already merged and to update workflow docs.
@ZBSingleton for markets that haven't been merged yet (and future workflows).

stepps00:
Related to: https://github.com/whosonfirst-data/whosonfirst-data/issues/673

nvkelso:
This is a fix for https://github.com/whosonfirst-data/whosonfirst-data/issues/643.

nvkelso:
Closed via https://github.com/whosonfirst-data/whosonfirst-data/issues/677

PR #677, opened 2017-02-27
Airport record updates - Campus and Neighbourhood placetypes

stepps00:
This PR fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/673 and https://github.com/whosonfirst-data/whosonfirst-data/issues/672.

Includes the following updates:
- Updated the geometries for Airport `campus` records that are equal to (or share boundaries with) their parent `neighbourhood` shape.
- Added `reversegeo` properties to the `campus` records
- Retained the existing woedb-sourced geometry in the alt-geometry slot.
- Added "District" to the end of each Airport neighbourhood record.

Each neighbourhood/campus update required a slightly different case; work is split into the three commits below.


data/102/532/251/102532251.geojson
```diff
@@ -71,7 +75,7 @@
         }
     ],
     "wof:id":102532251,
-    "wof:lastmodified":1486772328,
+    "wof:lastmodified":1488229998,
     "wof:name":"Portland International Airport",
```
nvkelso: This needs ` District` postfixed.

nvkelso: Oh, never mind.

nvkelso:
Will require re-PIPing for the campus geoms.

Issue #645, opened 2017-02-13
places with the name 'City'


missinglink:
the following wof records have the name 'City' which, albeit possible, are unlikely the correct name:

```
City|default|101938739
City|default|404516183
City|default|404549959
City|default|420781621
City|default|420782383
City|default|420784611
City|eng|101938739
City|eng|404516183
City|eng|404549959
```

stepps00:
thanks @missinglink - will take a look shortly.

stepps00:
**City|default|101938739**
  - This is the City Center area of Canberra, AUS. 
  - According to [Wikipedia](https://en.wikipedia.org/wiki/Civic,_Australian_Capital_Territory), " its official division name is City." 
  - No changes required.

**City|default|404516183**
  - This is a township in Missouri.
  - [Wikipedia](https://en.wikipedia.org/wiki/City_Township,_Barton_County,_Missouri) confirms the name is "City Township."
  - Since there are no placetype suffixes in the `wof:name`, no changes required.

**City|default|404549959** (see **101938739** above)
  - This is the City Center area of Canberra, AUS. 
  - According to [Wikipedia](https://en.wikipedia.org/wiki/Civic,_Australian_Capital_Territory), " its official division name is City." 
  - No changes required.

**City|default|420781621**
  - This _likely_ represents the "City of London", as its fairly close to the center London's city boundary. I can not find any information online about a "City"neighbourhood in London.
  - London neighbourhood updates are being tracked in issue #411.
  - Suggest either deprecating this record now or waiting for London neighbourhood work to be completed.

**City|default|420782383**
  - This is a district in Zurich.
  - According to sources ([1](https://en.wikipedia.org/wiki/Altstadt_(Z%C3%BCrich)#City), [2](https://www.inyourpocket.com/zurich/Zurichs-districts_71823f)), this is a valid name for an area of Zurich.
  - No changes required.

**City|default|420784611**
  - This _likely_ represents the Indre By area or City Center area of Copenhangen. I can not find any information online about a "City"neighbourhood in Copenhagen.
  - Suggest deprecating this record now.

nvkelso:
I’m fine deprecating 420784611.

For 420781621 there is technically several localities for “London”,
including one for the older “City of London”, so I’d leave it alone until
we get into it with #411.

missinglink:
wow, thanks for looking in to it!

yet another interesting quirk of the world we live in that makes geocoding tricky :)

[...]

PR #678, opened 2017-02-27
Deprecated bunk 'City' record in Copenhagen

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/645.

Deprecates the "City" record in Copenhagen, adds an `is_current` flag of `0`, and updates the `lastmodified` date.

Should be a simple merge to `master`, no PIP work required.


nvkelso:
This was a point geometry that wouldn't get PIP'd, so doesn't need to be un-PIP'd (can be merged right away).

Issue #7, opened 2015-08-13
Bassas da India isn't a dependency


nvkelso:
wof_id  iso name    parent_id
85632337    TF  Bassas da India 85633147    

data/856/323/37/85632337.geojson

Bassas da India mapunit of 85632549 French Southern and Antarctic Lands (that region doesn't really have administrative subdivisions, each island is shown as a map unit in Natural Earth).

nvkelso:
Ah, this should be disputed with Morocco, and just a tiny island between Madagascar and the African main continent.

stepps00:
#687 and #681 have been merged. [Bassas de India](https://whosonfirst.mapzen.com/spelunker/id/1108833329/) is now a `disputed` record, closing issue.

PR #681, opened 2017-03-08
Deprecated, superseded Bassas da India record

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/7.

- Deprecated [Bassas da India](https://whosonfirst.mapzen.com/spelunker/id/85632337/#3/-32.63/58.66) record, which is a dupe of the [French Southern and Antarctic Lands](https://whosonfirst.mapzen.com/spelunker/id/85632549/#3/-32.63/58.66) record.
- Added is_current property
- Tested out a new `edtf:superseded` date

Shouldn't require PIP work - no descendants under the Bassas da India record

PR #687, opened 2017-03-15
Updated Bassas da India, created new record

stepps00:
Replaces https://github.com/whosonfirst-data/whosonfirst-data/pull/681. 

- Superseded the Bassas da India record into a new Bassas da India record
- New Bassas da India record is a `disputed` placetype, with dual hierarchies between France and Madagascar
- New record does not include NE data for the French Antarctic Islands
- Gave `edtf:cessation` and `edtf:superseded` dates to the superseded (existing) record


stepps00:
Fixes #7.

nvkelso:
Over to @thisisaaronland. Don't think we have any venues here, but technically requires PIP work.

stepps00:
Metafiles will need to be regenerated, but I don't believe this will require any PIP work. I think we can merge as-is, unless I'm missing something here.

Issue #534, opened 2016-11-22
Add Australia regions


nvkelso:
From the new Australia data

nvkelso:
Partial work towards #212.

nvkelso:
This is a good way to test out alt-geom wrangling, related to #367.

Issue #218, opened 2016-03-03
Investigate official Australian boundary data


riordan:
- [x] postcode
- [x] region (in PR)
- [ ] localadmin
- [ ] locality
- [ ] neighbourhood (wards), is this usable?
- [ ] constituency

Feb 26 2016

https://www.psma.com.au/publications/news/psma-responds-australian-governments-release-g-naf-public

stepps00:
[Link](http://data.gov.au/dataset/psma-administrative-boundaries) to administrative data (CC BY 4.0)

rowanwins:
Let me know if you require any local input into deciphering these datasets @stepps00

stepps00:
**Notes, per placetype:**

Who's On First | PSMA | Note 
---|---|---
country | country | No action. No work to do at this placetype.
region | state boundaries | No action. In PR.
county |  local government areas | Update WOF features with newer 2016 data.
localadmin | N/A | Data available from ABS.
locality | locality | Only a portion of this PSMA data is in WOF. Check placetype, supersede work likely.
neighbourhood | N/A | Find data.
postalcode | postalcode | No action. Imported.
constituency | electoral district | Create new constituency records with electorcal district data.

nvkelso:
Let's look at this one together, sooner or later.

stepps00:
@stepps00 - Find names for all features that have a proposed placetype change. 
@nvkelso - Double check the approach is correct.

stepps00:
**Related:**
```
aus_adm0.shp:
    ogr2ogr -overwrite -f "ESRI Shapefile" -lco ENCODING="UTF-8" -sql "SELECT 'AUS' as qs_adm0_a3, 'Australia' as qs_adm0, 'adm0' as qs_level, 'Australia Census' as qs_source, 'AU' as qs_iso_cc, 'Australia' as qs_a0 FROM TR_2012_AUST" aus_adm0.shp ../nma/australia/1270055003_tr_2012_aust_shape/TR_2012_AUST.shp

    #dissolve-shapes.py -i aus_adm0.shp -o aus_adm0.shp -f qs_adm0_a3,qs_adm0,qs_level,qs_a0
    
aus_adm1.shp:
[...]

PR #688, opened 2017-03-17
New geometries and alt_geom records for Australia regions

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/534.

Using the PSMA data (linked in issue https://github.com/whosonfirst-data/whosonfirst-data/issues/218), this PR:

- Moved the existing Quattroshapes geometries to an alt_geom record for each region in Australia
- Updated the default record's geometry with the PSMA (`ausstat`) geometries
- Updated the `lbl:latitude` and `lbl:longitude` properties from Mapshaper values
- Updated the `src` property values

It looks like Quattroshapes sourced the larger region geometries from PSMA data to begin with, but the new geometries are slightly more detailed. Island region geometries are higher quality, though.


nvkelso:
New geoms = PIP'ing required.

Issue #692, opened 2017-03-20
Incorrect superseding value for Northeast PDX neighbourhood


stepps00:
The [neighbourhood record](https://whosonfirst.mapzen.com/spelunker/id/85893213/#15/45.5567/-122.6448) of "Northeast" in Portland has a `wof:superseded_by` value of "1108720843". This value should actually be "1108720843" to point to the new [Northeast macrohood](https://whosonfirst.mapzen.com/spelunker/id/1108720843/#12/45.5658/-122.5784) record.

nvkelso:
(This is related to Nat's neighbourhood curation import.)

PR #693, opened 2017-03-20
Updated superseded_by value for NE PDX

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/692.

- Updated the Northeast Portland nighbourhood's `wof:superseded_by` value to point to the correct record (Northeast macrohood - [1108720843](https://whosonfirst.mapzen.com/spelunker/id/1108720843/#12/45.5658/-122.5784))
- Ran through Exportify tool

Simple property update, should not require PIPing.


nvkelso:
Over to @thisisaaronland. This is property edit only, no PIP needed. Metafiles need to be re-generated.

This fix will unblock me for Nat's new neighbourhood import, too.

Issue #696, opened 2017-03-20
Upgrade Austrian region records' geometries


stepps00:
Following https://github.com/whosonfirst-data/whosonfirst-data/issues/546, the Austrian locality data discussed in that issue have a property that includes the name of the region that the locality falls within. 

Dissolve the locality data on the region name and update the Austrian region records with a new, more detailed geometry.

<img width="1003" alt="screen shot 2017-03-20 at 1 03 25 pm" src="https://cloud.githubusercontent.com/assets/18567700/24119216/a514ac90-0d6d-11e7-9a87-eadeb9d91d24.png">

stepps00:
Fixed.

PR #697, opened 2017-03-20
Updated geometries for Austrian region records, new alt-geoms

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/696.

- Updated `mz`-sourced geometries for the nine region records in Austria
- Nine new alt_geometry records that contain the original `quattroshapes` geometry
- Updated `lbl` properties from new Mapshaper values
- `src:geom` and `src:geom_alt` property updates; `mz` and `quattroshapes, respectively


nvkelso:
Over to @thisisaaronland. Will require re-PIPing (the shapes are different in detail along roads, but not general substance – same number of regions in and out).

PR #716, opened 2017-04-04
Nat 2017/4/4 neighbourhoods with moved feature geom (27 require PIP)

nvkelso:
This PR is limited to just the 27 features which were POINT only geometries where the label moved, which effectively moved the PIP geometry, requiring a re-PIPing.

PR #720, opened 2017-04-05
Nat 2017/4/4 neighbourhoods with moved feature geom (27)

nvkelso:
Replaces https://github.com/whosonfirst-data/whosonfirst-data/pull/716.


nvkelso:
Over to @thisisaaronland. These point geoms were moved, so requires PIPing.

PR #718, opened 2017-04-04
Nat 2017/4/4 neighbourhoods with new features (require PIP)

nvkelso:
-  6912 records

PR #722, opened 2017-04-05
Nat 2017/4/4b neighbourhoods new (requires PIP) 6912 records

nvkelso:
Replaces https://github.com/whosonfirst-data/whosonfirst-data/pull/718.


nvkelso:
(Since these are all points PIP only needs to be run to complete the hierarchy – there aren't any descendant venues to rip thru.)

Issue #719, opened 2017-04-05
WOF id 1108692869 missing from county bundle


trescube:
wof-county-latest.csv contains WOF id 1108692869 that has datapath `110/869/286/9/1108692869.geojson` but this file is missing from the county bundle.

nvkelso:
I suspect the following 5 features are recent **neighbourhood** updates that point to a new record that has a bad `wof:id`:

- data/858/461/69/85846169.geojson
- data/858/511/47/85851147.geojson
- data/858/659/71/85865971.geojson
- data/858/668/53/85866853.geojson
- data/858/874/21/85887421.geojson

I found them while importing Nat's neighbourhood curation.

stepps00:
@trescube - this issue should be resolved now that #723 was merged. Is this still an issue for Pelias?

trescube:
That fixed it, thanks!

PR #723, opened 2017-04-05
Fixed bunk supersedes ids, created one new feature

stepps00:
This PR updates all bunk `wof:superseded_by` and `wof:supersedes` values listed in #https://github.com/whosonfirst-data/whosonfirst-data/issues/719.

- Each invalid `wof:id` was replaced with the correct `wof:id` and features were run through the Exportify tool. 

- Feature `data/858/461/69/85846169.geojson` was incorrectly deprecated, so a new record was created (`data/110/895/148/3/1108951483.geojson`) and superseding work was done between the two.

This PR **does not** include updates to the bundles/meta files.


nvkelso:
Over to @thisisaaronland. No PIPing required, but metafiles will need to be generated.

PR #754, opened 2017-04-25
wof:population property, round one - Canada

### LINKED ISSUE #240: Expose `wof:population` property
The Pelias team has been pulling population data from the `gn:population` property and have just realized that we need to pull from `zs:pop10` for presumably neighborhoods.  Exposing a `wof:population` property would be great for an authoritative field and reducing complexity in our logic.  

Note from @nvkelso, with exact logic described in https://github.com/whosonfirst-data/whosonfirst-data/issues/240#issuecomment-294907374:

- No PIPing required as this will be a property only change, for all administrative records in the generic `whosonfirst-date` repo.
- No modification of existing properties, only addition of the following `common` properties. 
- [x] Add new `wof:population`
    - _a positive integer value_
    - _when not known don't set property_
- [x] Add new `src:population`
    - _a string from one of registered the wof:sources_
    - _should always be present when `wof:population` is present (e.g., if not known then default to `unknown`)_
- [x] Add new `src:population:date`
    - _a EDTF date, with default of `uuuu` when unknown – most of our current data is unknown dates._
    - _always present if there is `wof:population` value_
- [ ] Add new `src:population:method`
    - _not for this commit_
    - _string value of: `census`, `estimate`, etc_
- [x] Add new `wof:population_rank`
    - _a calculated integer when `wof:population` is present, but..._
    - _if no present `wof:population`, then okay for us to fill this in anyhow manually, which is how many features coming from Quattroshapes got their ranks._

**Question:** 

- What do we do with historic population time series?
- I propose punting that for another day, but here's an idea we talked about:

Optional data structure where `wof:population_timeseries` is an array, like:

```
'wof:population_timeseries': [
     { population: X1, source: Y1, date: Z1, method: A1, population_rank: B1}, 
     { population: X2, source: Y2, date: Z2, method: A2, population_rank: B2}, 
     { population: X3, source: Y3, date: Z3, method: A3, population_rank: B3}
}
```

stepps00:
This PR includes population work for all admin records in Canada, as a first pass to close #240.

Records include the following:

- If the record has no existing population property, the record was given a default `wof:population_rank` property with a value of `0`.
- If the record has an existing population property, the record was given:
  - A new `src:population` property that lists the source of the population figure.
  - A new `wof:population` property with an integer value equal to the existing population value in the record (priority of existing propertiyes in #240).
  - A new `wof:population_rank` property with an integer value, based on the `wof:population` figure.

**Script used to complete this work:** https://github.com/mapzen/whosonfirst-toolbox/blob/master/scripts/issue-240-wof_population.py

Follow-up PR(s) will include this work for **all** admin records in `whosonfirst-data`.


nvkelso:
I spot checked a bunch of these, including Vancouver and LGTM.

PR #762, opened 2017-04-28
New reversegeo:geometry prop for admin0 records

### LINKED ISSUE #224: Indicate preferred geom for reverse geocoding
We include alternate geometries with WOF, with the primary being a mix of Natural Earth and Quattroshapes. "Primary" in this case is for map display (just like the name property).

To support reverse geocoding (point in polygon, PIP) use cases, we should indicate which geometry we think is "best". Let's set all those to Quattroshapes for now.

Reverse geocoding would like the more detailed shape, and preferably the shape that extends out into the coastal waters (10 mile, 100 mile nautical limits) – both to capture the reality on the ground to capture the greatest territorial extent, and because that serves to simplify the shape, discarding tiny coastline details that otherwise bloat file size and decrease reverse geocoding PIP performance.

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/367. ~~Fixes #224~~

- Updated all admin0 records to include a new `reversegeo:polygon` property
- If the record contained "quattroshapes" as a value in the `src:geom_alt` property, a new `reversegeo:polygon` property was created, with a value of "quattroshapes"
- Otherwise, the record tdid not receive a new `reversegeo:polygon` property (the default geom is likely from quattroshapes already

**Script:** https://github.com/mapzen/whosonfirst-toolbox/blob/master/scripts/issues_367_reversegeo_polygon.py


nvkelso:
Can you add this to the wof-properties repo, please?

stepps00:
@nvkelso - New `reversegeo.md` file created here: https://github.com/whosonfirst/whosonfirst-properties/blob/master/properties/reversegeo.md

nvkelso:
@stepps00 If the country didn't have a change can you remove it from the PR, please? For instance https://github.com/whosonfirst-data/whosonfirst-data/pull/762/files#diff-d8601a3968de67c1e08ba1ecdf513182 didn't receive this new property, but it was still modified with new date.

stepps00:
Commit d561ea5 removed updates to records that did not receive a new `reversegeo` property. 
Commit 700f2a6 fixed all merge conflicts in branch.

nvkelso:
This doesn't require PIPing, but we should link this PR to WOF and Pelias work to ingest the specified `reversegeo:polygon` for point-in-polygon workflows. Also @thisisaaronland to comment on the new property name and value setup.

thisisaaronland:
It mostly seems fine although I don't understand the documentation here:

https://github.com/whosonfirst/whosonfirst-properties/blob/master/properties/reversegeo.md#polygon

The documentation lists this:

```
The reversegeo:polygon value is represented as {source} in the following filename expansion:

85922583-alt-{source}-extras.geojson
```

But it's not clear what `extras` is supposed to represent or what its relationship with reverse geocoding is.

It's also not clear why the `reversegeo:polygon:` property is truncated to just "quattroshapes". What happens when we generate display polygons from quattroshapes (or others) as a soutce?

Given all that I would expect the fully qualified filename (for a reversegeo poly) to be:

`85922583-alt-{source}-{function_or_purpose}.geojson` – for example `85922583-alt-quattroshapes-reversegeo.geojson`

For the sake of brevity we could decide that `reversegeo:polygon=quattroshapes` can be interpreted as `{function}:polygon={source}` but that should be explicitly documented, including the rules for how someone can automate determining that `reversegeo:polygon=quattroshapes-reversegeo` is _not_ the same thing and the expansion rules do not apply.

stepps00:
These are all valid points.. I think the best thing to do for this issue would be to:

  - Update the existing alt-geometry records to include the `"extras"` in the filename (change from `85922583-alt-quattroshapes.geojson` to `85922583-alt-quattroshapes-reversegeo.geojson`)
  - Change the default records' `reversegeo:polygon` property to include the full filename (change from `"quattroshapes"` to `"85922583-alt-quattroshapes-reversegeo.geojson"`)

To the last point, this rule should apply to any `{function}:polygon` property we implement later on. If we decide to include a `display:polygon` property, I'd expect the value (and alt-geom filename) to be:

  `85922583-alt-{source}-display.geojson`

nvkelso:
Display will be a little different in that we’ll support zooms and we
should recommend a cascade when data is sparse, but we can cross that
bridge later.

stepps00:
The two most recent commits take care of any merge conflicts in this branch.

This PR is a good candidate to test the pipeline tools; no PIP work is required for this PR.

nvkelso:
@stepps00 are you still planning on addressing https://github.com/whosonfirst-data/whosonfirst-data/pull/762#issuecomment-298438430?

nvkelso:
IRL, we discussed, for the 41 features:

`"reversegeo:polygon": "85922583-alt-quattroshapes.geojson"`

Without going thru the hoops of creating duplicate reversegeo alt geoms at this time. (If we're going to take the time it's more useful to do that for display geoms at different zooms).

thisisaaronland:
For the sake of pedantry that should probably be `{LABEL}:geometry` since the term "polygon" has a fixed and important meaning in GeoJSON (relative to multipolygons)

stepps00:
I'm open to `reversegeo:geometry`, as we use `geom` over `polygon` in current property names.

Most recent commit has **`reversegeo:geometry`** property.

nvkelso:
Fine by me. Over to @stepps00 and @dphiffer for new merge workflow.

dphiffer:
Hmm ... this pipeline skipped over the user confirmation step. But otherwise it worked.

PR #763, opened 2017-04-28
Updates to dependency hierarchies

### LINKED ISSUE #750: Clean up dependency hierarchies
Many dependency records have countries in their `wof:hierarchy`. This is causing countries to display _interesting_ descendant records. Examples:

- There are [83 region records](https://whosonfirst.mapzen.com/spelunker/id/85633793/descendants/?exclude=nullisland&placetype=region#2/39.2/-169.5) parented by the United States because dependencies like Guam and Samoa have region descendants
- There are [121 region records](https://whosonfirst.mapzen.com/spelunker/id/85633147/descendants/?exclude=nullisland&placetype=region#2/21.0/-178.1) parented by France because dependencies like Mayotte and Polynesia have region descendants

Hierarchies of a dependency record should only include the dependency (itself), its empire, and its continent. Dependency descendant records should inherit the same values (and not include countries).

This may also cause us to rework some of the country geometries. The geometry of [France](https://whosonfirst.mapzen.com/spelunker/id/85633147/#2/18.8/-3.0) ([related PR](https://github.com/whosonfirst-data/whosonfirst-data/pull/365)) includes some overseas dependencies, the [United States](https://whosonfirst.mapzen.com/spelunker/id/85633793/#1/65/0) does not.

stepps00:
Fixes #750 

- New Netherlands empire record
- Updated all dependency records as follows:
  - Removed `"country_id"` from the dependency hierarchy
  - Updated the `"wof:parent_id"` to be equal to the `"empire_id"` in the record's hierarchy

Script: https://github.com/mapzen/whosonfirst-toolbox/commit/2efbdc251ea928e23a6ebe47c44332133fce4357


stepps00:
@nvkelso - this PR only updates the hierarchies of dependency records. To fix the issues actually listed in #750 (more than 50 regions in the US, etc), the descendant records of these dependencies also need to be updated. 

Should this PR include updates to those descendant records as well?

stepps00:
Note: Update all descendant records before merge.

nvkelso:
We'll need to update dependency docs in https://github.com/whosonfirst/whosonfirst-placetypes/issues/6 to correct the original source of confusion. Pelias already has this right.

nvkelso:
@stepps00 Please either confirm with @thisisaaronland that PIPing will take care of fixing all children's hierarchies, or add to this PR to correct all children manually.

stepps00:
@nvkelso - I will confirm whether or not the hierarchy changes "trickle down" to descendants, but in the meantime I will add the descendant record updates to this PR.

nvkelso:
@thisisaaronland will the hierarchy changes in this PR for dependencies "trickle down" to their descendants or will @stepps00 need to add those manually as part of this PR?

thisisaaronland:
This PR will break the tooling – or more specifically the tooling won't know what to do with these commits – because `empire` is not listed as a possible parent of `dependency` here (and by extension all the tooling): 

https://github.com/whosonfirst/whosonfirst-placetypes/tree/master/placetypes

Or rather we change the definition of dependencies to _only_ be parented by empires.

nvkelso:
Issue https://github.com/whosonfirst/whosonfirst-placetypes/issues/6 covers updating the place types – is anything else needed besides that to make this work?

thisisaaronland:
I'll just need to make sure that the various tooling (Python libraries, etc.) have been updated before processing the PR but otherwise yes, it should be handled correctly.

Issue #767, opened 2017-05-02
Country records missing name:eng_x_preferred property


stepps00:
Per the [language preferred test](https://github.com/whosonfirst/whosonfirst-tests/blob/master/test_runs/test_lang_names.js), the following countries are missing a `name:eng_x_preferred` property:

```
Test failed - 85633267 (Liechtenstein) is missing a eng_x_preferred field. (placetype: country)
Test failed - 85633275 (Luxembourg) is missing a eng_x_preferred field. (placetype: country)
Test failed - 85633285 (Monaco) is missing a eng_x_preferred field. (placetype: country)
Test failed - 85632727 (Netherlands) is missing a eng_x_preferred field. (placetype: country)
Test failed - 85633147 (France) is missing a eng_x_preferred field. (placetype: country)
Test failed - 85632343 (Andorra) is missing a eng_x_preferred field. (placetype: country)
Test failed - 85632355 (Paraguay) is missing a eng_x_preferred field. (placetype: country)
Test failed - 85633735 (Portugal) is missing a eng_x_preferred field. (placetype: country)
Test failed - 85633763 (San Marino) is missing a eng_x_preferred field. (placetype: country)
Test failed - 85632647 (Togo) is missing a eng_x_preferred field. (placetype: country)
Test failed - 85632371 (Saint Pierre and Miquelon) is missing a eng_x_preferred field. (placetype: country)
Test failed - 85632511 (Uruguay) is missing a eng_x_preferred field. (placetype: country)
Test failed - 85632523 (Mayotte) is missing a eng_x_preferred field. (placetype: country)
Test failed - 85632765 (Sao Tome and Principe) is missing a eng_x_preferred field. (placetype: country)
Test failed - 85632599 (Nicaragua) is missing a eng_x_preferred field. (placetype: country)
Test failed - 85632185 (Sint Maarten) is missing a eng_x_preferred field. (placetype: country)
Test failed - 85632261 (Ecuador) is missing a eng_x_preferred field. (placetype: country)
Test failed - 85632281 (Angola) is missing a eng_x_preferred field. (placetype: country)
Test failed - 85632317 (Venezuela) is missing a eng_x_preferred field. (placetype: country)
Test failed - 85632325 (Chad) is missing a eng_x_preferred field. (placetype: country)
Test failed - 102312305 (United Nations) is missing a eng_x_preferred field. (placetype: country)
```

PR #769, opened 2017-05-02
Added missing wof:lang name props for country records

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/768.

This PR includes the following:

- New `name:tvl_x_preferred` prop for Tuvalu
- New `name:gil_x_preferred` prop for Kiribati (lang: Gilbertese)
- New `name:fre_x_preferred` prop for Saint Pierre and Miquelon
- New `name:zho_x_preferred` and `name:por_x_preferred` prop for Macao S.A.R.
- New `name:chi_x_preferred` prop, `name:eng_x_preferred` prop and `name:fre_x_preferred` prop for the United Nations

These names were "missing", i.e. these languages were in the `wof:lang` property for their respective country, but the preferred`name:` property for that language was missing. Languages come from Wikipedia/.

PR #770, opened 2017-05-02
Added missing eng_x_pref props to country records

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/767.

- New `name:eng_x_preferred` properties for all countries that do not current;y have a `name:eng_x_preferred` property. 
- In some cases, this name was already stored in the variant slot; n those cases, it was removed from the variant slot and just added as a preferred name. 
- Each new preferred name should equal the `wof:name`.


stepps00:
@nvkelso - pinging you for review on this issue and #769, these PRs don't need any PIP work done and will clear up the `admin0` name work we're doing.

nvkelso:
Over to @thisisaaronland, no PIP required.

Issue #776, opened 2017-05-03
Add placetype_alt for Piedmont, California


nvkelso:
Let's pretend it's a "neighborhood" of itself, surrounded by Oakland neighbourhoods.

PR #777, opened 2017-05-03
New placetype_alt property for Piedmont

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/776

- Added a new `wof:placetpye_alt` property for Piedmont, CA
- Exportified record


nvkelso:
Property edit only. Update metafile, no-PIP, merge :)

Issue #507, opened 2016-10-27
Add wof:lang property to county records of Puerto Rico


stepps00:
The 78 `county` records in Puerto Rico ([here](https://whosonfirst.mapzen.com/spelunker/id/85633729/descendants/?exclude=nullisland&placetype=county#9/18.2228/-66.7570)) should be given a new `wof:lang` field. The following should be added to each record:

```
wof:lang: "spa";
```

See: https://github.com/pelias/wof-pip-service/pull/62

Issue #511, opened 2016-11-03
Update correct wof:lang for Finnish regions


hannesj:
The following regions incorrectly specify `eng` in `wof:lang`, which should be removed.

```
Uusimaa region 85683067                   
Southwest Finland region 85683073         
Tavastia Proper region 85683077           
Kymenlaakso region 85683081               
Päijänne-Tavastia region 85683085         
Paijanne Tavastia region 85683091         
Pirkanmaa region 85683099                 
South Karelia region 85683103             
Southern Savonia region 85683109          
Southern Ostrobothnia region 85683113     
Central Finland region 85683117           
Ostrobothnia region 85683119              
Central Ostrobothnia region 85683125      
Northern Savonia region 85683131          
North Karelen region 85683133             
Kainuu region 85683137                    
Northern Ostrobothnia region 85683143     
Lapland region 85683147                   
```

Also, the two records for Päijänne-Tavastia (85683085 and 85683091) should be merged.

This is related to pelias/wof-pip-service#62

hannesj:
Information about the official languages for Finnish `localadmin`s can be found [here](https://en.wikipedia.org/wiki/List_of_municipalities_of_Finland_in_which_Finnish_is_not_the_sole_official_language).

stepps00:
```
Also, the two records for Päijänne-Tavastia (85683085 and 85683091) should be merged.
```

This is being taken care of in :https://github.com/whosonfirst-data/whosonfirst-data/pull/514

Issue #290, opened 2016-05-16
Backfill wof:lang_x_* properties


thisisaaronland:
Map wof:country to ISO2 country defined here:

https://github.com/openvenues/libpostal/blob/master/resources/language/countries/country_language.tsv

Note - we will still need to make ISO-639-2 to ISO-639-3 ourselves.

nvkelso:
And by region, but using IDs of mixed craziness: https://github.com/openvenues/libpostal/blob/master/resources/language/regional/adm1.tsv.

thisisaaronland:
Separate issue. Simple things, first.

thisisaaronland:
Conveniently, GeoPlanet did most of the work for us already:

```
python ./issue-290-countries.py ./issue-290-countries.json  /usr/local/data/whosonfirst-data/meta/wof-country-latest.csv 
missing wof:lang for 85632231
missing wof:lang for 85632295
missing wof:lang for 85632233
missing wof:lang for 85632161
missing wof:lang for 85632709
missing wof:lang for 85632329
missing wof:lang for 85632379
missing wof:lang for 85632315
missing wof:lang for 85632593
missing wof:lang for 85632607
missing wof:lang for 85633793
missing wof:lang for 85633259
missing wof:lang for 85632495
missing wof:lang for 85632423
missing wof:lang for 85632461
missing wof:lang for 85632469
missing wof:lang for 85632437
missing wof:lang for 85632413
has wof:lang 190 missing wof:lang 18
```

https://whosonfirst.mapzen.com/spelunker/id/85633143/ <-- wof:lang=["fin"]

thisisaaronland:
Note to self:

1. Apply updates from country records to regions - sudo make me a tool
2. Take a stab at updating regions per libpostal file above, in a separate issue
3. Apply updates to descendants of regions accordingly - see above inre: "a tool"

stepps00:
Note: per https://github.com/whosonfirst/whosonfirst-properties/blob/master/properties/wof.md, the `wof:lang` property is now deprecated. We should use this issue to track updating `wof:lang_x_official` and `wof:lang_x_spoken` instead.

stepps00:
Re-running the issue-290 scripts yields the following:

```
python .scripts/issue-290-countries.py .scripts/issue-290-countries.json ./meta/wof-country-latest.csv
missing wof:lang for 421166797
Warning 1: OGR_G_Area() called against non-surface geometry type.
[...]

PR #795, opened 2017-05-19
Added wof:lang props for PR counties

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/507, related to a merged Pelias issue.

PR updates include:

- New `wof:lang` prop equal to `['spa']`
- New `wof:lang_x_spoken` prop equal to `['spa', 'eng']`
- New `wof:lang_x_official` prop equal to `['spa', 'eng']`
- Updated `wof:lastmodified` date

Additional updates are tracked in #290 and #511 for other locations/records.


data/102/080/271/102080271.geojson
```diff
@@ -83,7 +83,18 @@
     ],
     "wof:id":102080271,
     "wof:label":"Municipio de Manat\u00ed",
-    "wof:lastmodified":1475267157,
+    "wof:lang":[
+        "spa"
+    ],
+    "wof:lang_x_official":[
+        "spa",
+        "eng"
```
nvkelso: I don't see English offered on their website:

- http://www.manati.pr

nvkelso:
This Wikipedia article gives a good overview of language usage (official, spoken) in PR:

- https://en.wikipedia.org/wiki/English_language_in_Puerto_Rico

Summary:

- Across PR at a top level there are 2 official languages. 
- In practice some government functions are in English, some high schools are in English (either a foreign language class or most private school instruction). 
- In practice most people speak Spanish, especially away from the capital.

data/102/080/271/102080271.geojson
nvkelso: _**for all:**_ It's technically not wrong to list `eng` here as _some_ official business and spoken language is in English, but dominant is Spanish (as recorded here with `"wof:lang" = "spa"`).

It is weird (but probably not technically wrong) to not include an English preferred name.

nvkelso: Something like: `Manati Municipality`.

stepps00:
This was one of the websites I referenced when updating the language properties.. I chose to include English to err on the side of including more information and, like you pointed out, the 

`"...Puerto Rican government has declared English an official language on several occasions while removing it from that status on as many occasions. English remains an official language in the Commonwealth."`

I think the best approach here is to include English and Spanish, add add preferred names for all records in this PR (something like [this](https://github.com/whosonfirst-data/whosonfirst-data/pull/795#discussion_r117996736)).

stepps00:
Most recent commit addressed all comments in this PR.

All records received:
- A new `name:eng_x_preferred` property equal to the record's `wof:name`
- A new `name:eng_x_variant` property equal to the record's `wof:name` + ' Municipality'

This workflow is consistent with past `name` property updates.

Issue #180, opened 2015-12-16
Upgrade neighbourhoods in Buenos Aires


meetar:
According to our resident expert in Buenos Aires neighborhood topology, this neighborhood should only have the southernmost shape, and the other two polygons may belong to another neighborhood – or maybe the people who caused the northern areas to be included in this neighborhood were lost, and this may be a cry for help:

https://github.com/whosonfirst/whosonfirst-data/blob/master/data/857/643/11/85764311.geojson

______________________
_(stepps00: edited 02-02-2017)_

### tl;dr

Status:
- [x] Propose **sources**
- [x] Establish **concordances** between WOF and source
- [x] **Upgrade / add** WOF shapes
- [x] **Add new or missing** features to WOF
- [x] **Deprecate / supersede** WOF features
- [x] **Upgrade** locality record's shape
- [x] **Draft review**
- [ ] **Draft revisions** based on review comments
- [ ] **Final review**
- [ ] **Final revisions** based on review comments
- [ ] **Import** into Who's On First

### Details

Sourcing:
- Permissive license?
- Add new src to Who's On First

Upgrade / add:
- `macrohood`
- `neighbourhood`
- `microhood`

Add new or missing features to WOF
- `macrohood`
- `neighbourhood`
- `microhood`

Deprecate / supersede:
- `macrohood`
- `neighbourhood`
- `microhood`

nvkelso:
See also: https://whosonfirst.mapzen.com/spelunker/id/85764311/

<img width="1184" alt="screen shot 2015-12-16 at 14 33 55" src="https://cloud.githubusercontent.com/assets/853051/11855928/188840dc-a402-11e5-8d4f-d5f26e826d5f.png">

stepps00:
Pulled all neighbourhood records in Buenos Aires, the same issue is occurring with the following records:

https://whosonfirst.mapzen.com/spelunker/id/85764025
https://whosonfirst.mapzen.com/spelunker/id/85764079
https://whosonfirst.mapzen.com/spelunker/id/85764109
https://whosonfirst.mapzen.com/spelunker/id/85764113
https://whosonfirst.mapzen.com/spelunker/id/85764129
https://whosonfirst.mapzen.com/spelunker/id/85764135
https://whosonfirst.mapzen.com/spelunker/id/85764183
https://whosonfirst.mapzen.com/spelunker/id/85764185
https://whosonfirst.mapzen.com/spelunker/id/85764225
[...]

PR #660, opened 2017-02-17
Mesoshapes - Admin1 updates, round 1

stepps00:
This PR includes updated `admin1` features for ~100 countries. Several commits have comments specific to that country or workflow. This PR includes the following:

- New and updated region and macroregion records 
- Updated `hasc:id` concordance for all features (for future Statoids join)

This PR does not include updated metafiles. Placing "(WIP)" in the title because there is likely additional work to do after review. 

This PR should fix the following issues: #628, #497, #486.

PR #816, opened 2017-06-06
Buenos aires neighborhoods

mfogel:
Updated geometry and hierarchy for Buenos Aires neighborhoods, boroughs, locality and region. See #180 for discussion.

This PR affects 83 records:

* adds no new records
* makes BA region record (85668081) coterminous with locality record (421174285). Updated geometry for both.
* deprecates the 15 existing county records for the Comunas and supersedes them with 15 new borough records. The new borough records have updated geometries:
  * 1108717815 -> 1118717815
  * 1108717817 -> 1118717817
  * 1108717819 -> 1118717819
  * 1108717823 -> 1118717823
  * 1108717825 -> 1118717825
  * 1108717827 -> 1118717827
  * 1108717829 -> 1118717829
  * 1108717831 -> 1118717831
  * 1108717833 -> 1118717833
  * 1108717835 -> 1118717835
  * 1108717837 -> 1118717837
  * 1108717841 -> 1118717841
  * 1108717843 -> 1118717843
  * 1108717845 -> 1118717845
  * 1108717847 -> 1118717847
* updates the geometry for 48 neighborhood records: (421174889,421191195,85764025,85764035,85764055,85764061,85764071,85764073,85764079,85764099,85764103,85764109,85764117,85764119,85764125,85764133,85764149,85764193,85764205,85764209,85764225,85764227,85764235,85764239,85764243,85764247,85764251,85764263,85764271,85764273,85764275,85764281,85764287,85764289,85764301,85764307,85764311,85764315,85764319,85764325,85764329,85764333,85764347,85764349,85765607,85765611,85765613,85765619)
* deprecates 3 duplicate neighborhood records, and supersedes them with their better counterparts:
  * 420781697 -> 421191195 
  * 85764277 -> 85764347
  * 85767083 -> 85764227

I made updates to the properties of these records for which I found [docs](https://github.com/whosonfirst/whosonfirst-properties/). The one exception is the `wof:geomhash` tag which I've just left as the empty string for each record that received an updated geometry. My understanding is that this tag will be set as part of the import process.

The new source `ar-caba` is included in whosonfirst/whosonfirst-sources#75


nvkelso:
I'll let @stepps00 comment in more detail, but one question right away:

> deprecates the 15 existing county records for the Comunas and supersedes them with 15 new borough records. The new borough records have updated geometries:

In New York City we allow both `county` and `borough` records mostly because they happen to have different names but are essentially the same "things". That and `county` is a "common" placetype" where borough is not. 

For BA, we imported `county` features as part of our mesoshapes admin 2 sprint (the files this PR modifies). ¿Should we still have `county` records that are marked coterminous with their `borough` records, or should the `borough` records just get a `placetype_alt` of `county`?

Are there any properties we should bring in from the source?

stepps00:
I think the questions about `county` and `borough` placetypes above are worth considering - its also preferred that we don't have any gaps in the `county` placetype in Argentina, if possible.

I would like to take a closer look that this new data this week before passing off as an import task, but I have two questions after looking things over.

- Right now, it looks like Brooklyn Integers ([www.brooklynintegers.com](www.brooklynintegers.com)) is minting ids that begin with `110`, though some of your new `wof:id` values begin with `111`. How did you mint the new `wof:id` values in this PR?

- I noticed a few cases where the following property is included:
```
"src:geom_alt":["quattroshapes"],
```
I would expect:

```
"src:geom_alt":[
    "quattroshapes"
  ],
```

I suspect these features were not run through the Exportify tool, is this true? 

Either way, I will checkout this branch before import and ensure all the GeoJSON files are valid. Also, if anything in our documentation was unclear during your update process, please let us know!

mfogel:
I added some commits to address some of the issues mentioned. Other stuff:

Integers: I didn't realize Brooklyn integers was returning a monotonic increasing sequence... I thought the only requirement was to avoid collisions. I will add another commit tomorrow changing the new IDs to match what the brooklyn int api returns for me.

exportify script: somehow I missed the existence of this script. It's very helpful! I had actually written a handful of short python and shell scripts myself that did most of what exportify does. I would suggest, in the [docs](https://github.com/whosonfirst/py-mapzen-whosonfirst-export/blob/master/scripts/wof-exportify#L18-L30) for exportify, to add more explanation as to what a 'source' is... wasn't clear to me until I read the code.

I've left geomhashes still as the empty string - the exportify script, when I run it on my machine against existing, unchanged WOF records, gives a new and different geomhash. Hence I don't trust it.

mfogel:
Ok, aside from the empty `wof:geomhash` tags and the `county` vs. `borough` decision, I believe I've addressed all the comments above.

Re: `county` vs. `borough` for the Comunas. For how I'm using the WOF data, it doesn't matter which placetype is chosen here. All that matters is that the more accurate geometries in this PR are accepted. That being said, to me, it seems more natural to use `borough` rather than `county` per the reasons listed in #180 - namely setting 15 different hierarchies for the `locality` record for Buenos Aires seems like a lot of scaffolding for a relationship that doesn't exist in the real world. The city of BA does not report to the Comunas - it is the other way around. That being said, I can see how it would be nice to have complete `county` coverage of Argentina.

stepps00:
Thanks for addressing the comments, @mfogel - I will checkout the branch this afternoon, run each file through our Exportify tool to format and verify each record, and poke around to make sure all edits are ready for import.

I will let you know if I have any additional questions or comments.. thanks again!

mfogel:
It sounds like changing from `county` to `borough` might have been a reach to far with this PR... would this be easier to merge if I undo that change, and leave the 15 comunas as `county` for now?

stepps00:
Hey @mfogel - sorry for the delay! I think the proposed changes are OK as-is. I'll give the PR a thorough (re)review this week and let you know if anything sticks out. Thanks for the patience.

nvkelso:
Add `"placetype_alt": "county"` to the boroughs and then I'll be 💯.

mfogel:
The merge into master of #660 caused some merge conflicts with this PR to appear. I've gone ahead and merged master into this branch and resolved the conflicts on this side. The only non-trivial conflict was that 71514bc1ad23f63a61a75d49019b12e41097fdee replaced the quattroshapes geometry with a meso geometry for the `region` record for Buenos Aires. The meso geometry is certainly better, but the geometry released by the city of BA is better still, so in this PR I've moved the meso geometry to alternate as well, just like that of the quattroshapes.

I've also used the [wof-hash.go](https://github.com/whosonfirst/go-whosonfirst-utils/blob/master/cmd/wof-hash.go) utility to fill in `wof:geomhash` values for records with a new or changed geometry in this PR. The python version of this utility still gives me a different/wrong value.

And the borough records are now set up with `county` as a `placetype_alt`.

Let me know if there's anything more I can do to help this PR. Cheers.

stepps00:
This is fantastic, thank you for all the work in Buenos Aires @mfogel.. I will let you know of any questions I have before merging this branch.

stepps00:
Hey @mfogel - I've copied your branch and created a new PR with your work here: #881.

I'm going to close this issue and complete additional QA on the changes in that branch.

Issue #817, opened 2017-06-06
"name:ru_x_preferred_x_variant" (and friends) properties


thisisaaronland:
What are these? Why are the language codes two characters? What are they `x_preferred_x_variant` ? How can they be `x_preferred_x_variant` ? Are there other instances like this, besides `ru_*` ?

How did this happen? This is all spelled out in detail here: https://github.com/whosonfirst/whosonfirst-names/

stepps00:
Unique `name:*` property mappings:

```
        "name:ba_x_preferred_x_variant" :  "name:bak_x_preferred",
        "name:ba_x_variant"             :  "name:bak_x_variant",
        "name:baq.p_x_preferred"        :  "name:baq_x_preferred",
        "name:bu_x_preferred_x_variant" :  "name:bul_x_preferred",
        "name:ca_x_preferred_x_variant" :  "name:cat_x_preferred",
        "name:ca_x_variant"             :  "name:cat_x_variant",
        "name:cz_x_preferred_x_variant" :  "name:cze_x_preferred",
        "name:cz_x_variant"             :  "name:cze_x_variant",
        "name:du_x_preferred_x_variant" :  "name:dut_x_preferred",
        "name:du_x_variant"             :  "name:dut_x_variant",
        "name:en_x_colloquial_x_variant":  "name:eng_x_colloquial",
        "name:en_x_preferred_x_variant" :  "name:eng_x_preferred",
        "name:en_x_variant"             :  "name:eng_x_variant",
        "name:fe_x_variant"             :  "name:fre_x_variant",
        "name:fi_x_preferred_x_variant" :  "name:fin_x_preferred",
        "name:fi_x_variant"             :  "name:fin_x_variant",
        "name:fr_x_preferred_x_variant" :  "name:fra_x_preferred",
        "name:fr_x_variant"             :  "name:fra_x_variant",
        "name:ge_x_preferred_x_variant" :  "name:ger_x_preferred",
        "name:ge_x_variant"             :  "name:ger_x_variant",
        "name:gr_x_preferred_x_variant" :  "name:ell_x_preferred",
        "name:gr_x_variant"             :  "name:ell_x_variant",
        "name:hr_x_preferred_x_variant" :  "name:hrv_x_preferred",
        "name:hu_x_preferred_x_variant" :  "name:hun_x_preferred",
        "name:hu_x_variant"             :  "name:hun_x_variant",
[...]

PR #821, opened 2017-06-09
Remapped bad name:* property values

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/817.

Updated records to remove bad `name:*` properties listed in issue.

If the record already had the correct `name:*` property that the bad `name:*` property should have been, the name value was appended to the existing property and the bad `name:*` proeprty was removed.

If a valid `name:` property did not exist, the `name:*` property was renamed.

Additionally, the last commit removes any duplicate name values in any updated property (not a common occurrence, but it did affect a few dozen records).

This PR can be committed without any PIP work.


nvkelso:
LGTM so far :)

stepps00:
Most recent commit takes care of any duplicate values that snuck into a single `name:*` property. 

Examples:

`data/858/630/05/85863005.geojson`
`data/858/630/01/85863001.geojson`
`data/890/537/237/890537237.geojson`

Note:
```
if props.has_key(v):
  name_prop = props[v]
  props[v] = list(set(name_prop))
```

nvkelso:
@stepps00 I'm still seeing a few strange values like:

- `name:eng.p_x_preferred`
  - https://raw.githubusercontent.com/whosonfirst-data/whosonfirst-data/stepps00/name-prop-updates/data/858/630/05/85863005.geojson
  - https://raw.githubusercontent.com/whosonfirst-data/whosonfirst-data/stepps00/name-prop-updates/data/858/630/01/85863001.geojson

stepps00:
@nvkelso - good catch.. somehow `name:eng.p_x_preferred` was not added to my lookup list and was not updated in records. I'll take a second look at any `name:*` property that includes `.p` and update in a new commit.

stepps00:
@nvkelso - it looks like some of the bad `name:*` props with `.p` snuck through, but the last three commits should have taken care of _all_ bad `name:*` props across all records (examples linked above should be fixed).

PR is ready for another review.

nvkelso:
Over to @thisisaaronland, property edits only for names, no PIP required.

Issue #641, opened 2017-02-08
Add unlc:subdivision to wof:concordances property


stepps00:
Many features, like [Calvados, FR](https://whosonfirst.mapzen.com/spelunker/id/85683579/#9/49.0919/-0.3567) have a `unlc:subdivision` property. This property value should instead be stored in the `wof:concordances` property.

Propose a PR with all affected features that adds this field to the `wof:concordances`.

stepps00:
@thisisaaronland - I think you added this property to WOF features, am I understanding this right?

thisisaaronland:
I did. It may or may not be treated as a concordance but it should not otherwise be removed, because this: https://github.com/whosonfirst-data/whosonfirst-data/issues/221

PR #822, opened 2017-06-10
Added unlc concordance values

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/641.

For any feature that current has a `unlc:subdivision` property, a `unlc:id` concordance value was appended to the record's `wof:concordances` list.

As commented in the issue, the original `unlc:subdivision` property was kept in each record.


stepps00:
Fix `data/136/251/273/136251273.geojson` before merge (`misc:*` props).

data/136/251/273/136251273.geojson
```diff
@@ -13,6 +13,25 @@
     "iso:subdivision":"CA-QC",
     "lbl:latitude":52.03334,
     "lbl:longitude":-72.765013,
+    "misc:qs_a0":"Canada",
```
nvkelso: This should be `qs:a0` not `misc:qs_a0`.

nvkelso:
@stepps00 why `"unlc:id"` instead of `"unlc:subdivision"`?

```
    "unlc:subdivision":"AD-04",
     "wk:wordcount":993,
     "wof:belongsto":[
         102191581,
         85632343
     ],
     "wof:breaches":[],	
     "wof:concordances":{	
          "gn:id":3040131,	
          "gp:id":20070555,	
          "qs:id":1102991,	
          "unlc:id":"AD-04",
          "wd:id":"Q24276",	
          "wk:page":"La Massana"
      },
```

Other than that I spot checked these and they look good (except https://github.com/whosonfirst-data/whosonfirst-data/pull/822#discussion_r121244445).

data/136/251/273/136251273.geojson
stepps00: Yeah, that's a bug in my callback script. It only affects the `136251273.geojson ` record - I've noted to fix it in an upcoing commit.

stepps00:
@nvkelso - I wanted the concordance to be standardized like other concordance values, though I have found `fips:code` in the concordances list. Would you expect this to be a `unlc:subdivision` concordance instead of `unlc:id`?

nvkelso:
@stepps00 Regarding https://github.com/whosonfirst-data/whosonfirst-data/pull/822#issuecomment-307531059, I'd expect it to match the `id` value given in the source JSON here: https://github.com/whosonfirst/whosonfirst-sources/blob/master/sources/unlocode.json#L6. In this case they _do_ match – but it's not consistent with `unlc:subdivision` that we use elsewhere. 

Looking at the upstream source metadata at http://www.unece.org/cefact/codesfortrade/codes_index.html it looks like all follow the `CC-az1` regardless of subdivision (WOF: `region`) or port (WOF: `venue` or `campus`) or other feature class so I think using `id` here is okay, but confusing since the source doesn't concatenate them together (instead they'd have `CC` in one column and `az1` in the next. Deserves a note in our https://github.com/whosonfirst/whosonfirst-sources/blob/master/sources/unlocode.json as to how we construct the `id`. Suggest adding something like:

```
WOF concatenate the UN/LOCODE 2 character country code and 3 character
feature code with a `-` deliminator and name the result the `id`.
```

stepps00:
I've updated the `UNLOCODE` source in this commit: https://github.com/whosonfirst/whosonfirst-sources/commit/e37ca85dba0c3e5983cfa98a5cf21ce269e15c83.

I will push a new commit to this branch to change the existing `unlc:subdivision` concordance to `unlc:id` shortly.

stepps00:
@nvkelso - New `unlc:id` property should be included in all records, as well as `qs:` fixes for the record of Quebec, and a note in the UNLOCODE source file.

Ready for another review.

stepps00:
The two most recent commits take care of any merge conflicts in this branch.

This PR is a good candidate to test the pipeline tools; no PIP work is required for this PR.

nvkelso:
Or rather @stepps00 wants to merge it with @dphiffer using new tool @dphiffer built.

stepps00:
Closing in favor of #882.

Issue #612, opened 2017-01-12
Dependency Abbreviations


trescube:
We're trying to support abbreviations for dependencies in Pelias and encountering some inconsistent behavior with their usage.  For example:

* [Faroe Islands](https://whosonfirst.mapzen.com/spelunker/id/85633153) -- no `wof:abbreviation`, `wof:country` is `DK` (the Faroe Islands parent country)
* [Puerto Rico](https://whosonfirst.mapzen.com/spelunker/id/85633729) -- both `wof:abbreviation` and `wof:country` are `PR`
* [Bermuda](https://whosonfirst.mapzen.com/spelunker/id/85632731) -- no `wof:abbreviation`, `wof:country` is `BM`

Is this just a matter of curating the data to standardize on either `wof:abbreviation` or `wof:country` or are their nuances to particular dependencies that we are unaware of?

nvkelso:
Different people mean different things by “abbreviation”, can you provide
examples of expected output?

I hear abbreviation and I think of:

   - Faroe Islands
<https://whosonfirst.mapzen.com/spelunker/id/85633153> — Far.
   Is.
   - Puerto Rico <https://whosonfirst.mapzen.com/spelunker/id/85633729> —
   P.R.
   - Bermuda <https://whosonfirst.mapzen.com/spelunker/id/85632731> — Berm.

But `wof:abbreviation` may be an oops.

I’d expect it to be language dependent, something more like:

For example (see eng_a below):

{
    "wof:name": "Montreal",
    "wof:lang": [ "eng", "fre" ],
    "name:eng_p": "Montreal",
    "name:eng_a": "YMQ",
    "name:fre_p": "Montréal",
    "name:kor_p": "몬트리올",
}

trescube:
Well, countries have iso2 values in their `wof:country` fields and dependencies seem to all (or most) have [iso2 values](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2), too.  US states all have the standard 2 character `wof:abbreviation`.  So I guess I'm asking what the expected behavior for dependency abbreviations:  "which field should it be in" and "is the value for itself of it's parent"?

nvkelso:
Are you trying to use the "abbreviation" to build a text label or to support search faceting?

trescube:
[...]

Issue #221, opened 2016-03-05
Finish up UN/LOCODE import


thisisaaronland:
1864 remaining regions to sort out (either because there are no matches or multiple matches in the spelunker ES index). List attached.

thisisaaronland:
It's more like 1511 regions to be sorted, because of wacky dump. Revised dump files (1220 regions not found, 291 regions with multiple results) attached.

[wof-unlocode-subdivisions-todo-multiple.csv.txt](https://github.com/whosonfirst/whosonfirst-data/files/160024/wof-unlocode-subdivisions-todo-multiple.csv.txt)
[wof-unlocode-subdivisions-todo-notfound.csv.txt](https://github.com/whosonfirst/whosonfirst-data/files/160025/wof-unlocode-subdivisions-todo-notfound.csv.txt)

sabas:
I noticed you're using a csv with incorrect characters, I have made a more accurate version you can find here https://github.com/datasets/un-locode

Did you map also the single locations or you had done only the subdivisions?

thisisaaronland:
Oh, that's great - thanks.

Right now the goal is simply to assign the correct UN-LOCODE subdivision to their respective `region` in Who's On First and then to propagate that property down to all its descendants.

The idea being that some collections of data (venues, specifically) may need to be chunked in to multiple smaller (Git) repositories, at least for as long as we continue to use Git. The naming convention at its most granular will be `whosonfirst-data` + "-" + `PLACETYPE + "-" + `ISO2 country code` + "-" + `UN-LOCODE subdivision`.

sabas:
@thisisaaronland do you plan to add in the descendants also the five letter code ?

thisisaaronland:
By 5-letter code do you mean ISO2 country code + UN-LOCODE subdivision code, or something else?

sabas:
Country + Location from the code list, example
https://whosonfirst.mapzen.com/spelunker/id/101940061/#14/-37.1925/145.7053
would have the unlc:code = AUAVA per the codelist  https://raw.github.com/datasets/un-locode/master/data/code-list.csv

If you could drop me an email I could explain better :)

thisisaaronland:
Apologies for the slow reply. Subdivision codes will include their parent country code. For example:

```
 'unlc:subdivision': 'US-WV',
```

stepps00:
[...]

Issue #211, opened 2016-02-17
Records with mismatched countries


trescube:
Attached is a file containing WOF IDs that have an `iso2` property that doesn't match the country in the hierarchy.  The file format has 3 fields:

1.  WOF ID
2.  `iso2` property value
3.  hierarchy-referenced country `iso2` property

An example entry is:

`85632199,BL,FR`

This should be interpreted to mean that WOF ID `85632199` has an `iso2` property value of `BL` but it's hierarchy references country `FR`.  

Please note that WOF ID `85681265` has literal `-1` as it's `iso2` property.  

[mismatched_countries.txt](https://github.com/whosonfirst/whosonfirst-data/files/133646/mismatched_countries.txt)

nvkelso:
@trescube Thanks for creating this list! Some of these are fixes over Quattroshapes (which hand coded some country codes incorrectly), and some of these are errors probably in how the PIP was setup using bad math "centroids" that weren't actually in the feature, and some are just funky. 

Can you confirm where the (2) iso2 property value is coming from versus the (3) hierarchy-referenced country iso2 property? 

For the example, `85632199`, Saint Barthelemy, a dependency, WOF gives:

```
iso
country	BL
parent	FR
```

Which is probably an unfortunate naming, as `country` I think is meaning `admin-0` inclusive of country and dependency, and `parent` here refers to `empire`. @thisisaaronland to confirm.

Similarly:

```
wof
country	BL
```

Looks like the "empire" of France hasn't been fully modeled yet, though?

* https://whosonfirst.mapzen.com/spelunker/id/136253037/

The -1 literal is valid WOF'ism for weird stuff:

* https://whosonfirst.mapzen.com/spelunker/id/85681265/

trescube:
Correct, the iso2 property I'm referring to is coming from `iso:country`.  We actually use iso3 for country code in our WOF importer but since that data isn't available (at least not easily) we use an external library to convert iso2->iso3.  So far it's just been easier to use the `iso2:country` field in the record but I suppose we could use that field from the country hierarchy element.  Which approach is advisable?

trescube:
This is especially problematic for the 5,785 entries like:

`101878271,UK,UA`

Where the `iso2` value is UK whereas the correct value is UA.

thisisaaronland:
Interesting. This appears to have been part of a batch update in September, specifically this one where 250K records were updated all at once.

[...]

PR #823, opened 2017-06-12
Various: Added wof:abbreviation prop, verified all country codes

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/612.

This PR cleans up the country codes for `dependency` records and adds a new `wof:abbreviation` property. Notes about updated properties:

- The `iso:country` property value is the iso code that represents the dependency.
- The `iso:parent` property value is the iso code that represents the dependency's parent country.
- The `wof:abbreviation` property value is the iso code that represents the dependency.
- The `wof:country` property value is the iso code that represents the dependency's parent country.

**Example for Falkland ~~Faroe~~ Islands:**

```
    "iso:country":"FK",
    "iso:parent":"GB",
    "wof:abbreviation":"FK",
    "wof:country":"GB",
```

There was some confusion about the dependency country codes in the past, but this should hopefully clear up the dependency properties.

**Related:** https://github.com/whosonfirst-data/whosonfirst-data/pull/453


stepps00:
paging @nvkelso for review (non-PIP PR).

nvkelso:
I want to hit pause for a minute before reviewing the changes in this PR...

Right now we have a lot of properties that seem to all be doing the same things but in slightly different ways. It's not clear to me what Who's On First hopes to accomplish with all these properties, or how they're used by downstream consumers.

## My understanding

- `wof:country`
    - **common** property
    - **values:** uppercase **2-char [ISO_3166-1](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) code**, but not limited to those
        - WOF recognizes more countries than ISO, but we should keep in the reserved XX namespace when divergent
    - **what:** represents the `country` **or** `dependency` (ISO doesn't give `disputed` areas codes)
        - This is based on https://github.com/whosonfirst/whosonfirst-placetypes/issues/6 being implemented
        - Discrepancy in practice:
             - Nathaniel: use for 2 char code of `wof:hierarchy`'s `country_id`'s "country" code and when that's not present than `wof:hierarchy`'s `dependency_id`'s "country" code.
             - Aaron: ¿use for 2 char code of `wof:hierarchy`'s `country_id`'s "country" code and when that's not present than `wof:hierarchy`'s `dependency_id`'s **parent's** "country" code?
    - **prevalence:** present on all features of all placetypes (including venues)
    - **uses:**
       - **wof data repos:** for splitting venues into country repos
       - **spelunker:** n/a?
       - **pelias:** for constructing labels, faceting search results by "country"?
- `wof:abbreviation`
    - ¿**common-optional** property?
        - ¿represents the dependency or region _when not the same as `wof:country`_?
        - Originally added for region "abbreviations"
    - **values:** ¿uppercase **2-char [ISO-3166-2](https://en.wikipedia.org/wiki/ISO_3166-2) code**?
    - **prevalence:** ¿are these just on countries, dependencies, and regions? Or on all placetypes when their parent?
    - **confusion:** ¿how is this different than the UN location code for subdivisions? 
      - e.g.: `'unlc:subdivision': 'US-WV'`, and https://github.com/whosonfirst-data/whosonfirst-data/issues/221
      - is `'unlc:subdivision' being used to segment WOF data repos? if so, should that be a top-level wof property? I think we've moved to being explicate about this with `wof:repo` property and `unlc:subdivision` has been left on the side of the road.
    - **uses:**
       - **wof data repos:** not used / maybe a variant of `unlc:subdivision`?
       - **spelunker:** not used
       - **pelias:** for constructing labels, faceting search "country" results? for dependencies **and** regions like Puerto Rico (PR) and New York (NY)
- `iso:country` raw ISO code that represents the dependency, country, or region
    - **common** property
    - a 2-char [ISO_3166-1](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2) code
    - **prevalence:** present on all features of all placetypes (including venues)
    - **uses:**
       - **wof data repos:** n/a
       - **spelunker:** search result list displays "country" name derived from ISO code?
          - should probably be `wof:country` instead? 
       - **pelias:** for constructing labels, faceting search results?
           - but when does this win over `wof:country`?
- `iso:parent` WOF chosen ISO code that represents the dependency's or region's "parent" ISO country.
    - ¿**common-optional** property?
    - ¿DO ANY FEATURES HAVE THIS NOW?
       - if not, let's drop this property.
    - **uses:**
       - **wof data repos:** not used
       - **spelunker:** not used
       - **pelias:** not used

## My preference is option 1 below

- `iso:country` is always what ISO 3166-1 says it is (e.g.: United States)
- `wof:country` is _mostly_ what ISO 3166-1 says it is, except when:
    - we have a country ISO doesn't have (we'll invent "legal" XX code, e.g. Kosovo)
    - dependency's keep their ISO 3166-1 code (we don't set it to the empire's home country's 3166-1 code (Puerto Rico)
    - ISO incorrectly gave a region a country code, we'll replace with that region's country's 3166-1 code (French Guyana)
- `wof:abbreviation` I think this relates more to `wof:name` than `wof:country`
    - we _do_ allow prefered localized name's like: `name:eng_x_preferred` of any character length
    - we _should_ allow abbreviated localized name's like: `name:eng_x_abbreviation` of any character length (mostly 2 char, up to 6 char), see https://github.com/whosonfirst/whosonfirst-names#a-short-history, but don't seem to have any populated values, and we need a default to fall back to.
    - if Pelias needs a 2-char region code we can always add `iso:region` from ISO 3166-2, and since that can sometimes be numerical we could also allow `wof:region` on those? Suggest they would only be on `region` placetype records and not propagate down (to venues or other descendants). 

## Examples

It seems like we mostly do a mix of **options 1** and **2** now for dependencies, and @stepps00 proposed **option 5** in this PR (with `iso:parent`) – and I get super confused / less is better.

1. **United States**
  - **option 1:** it's a valid ISO 3166-1 "country", and WOF think's it's an admin-0 `country`, profit!
    - wof:country: `US`
    - wof:abbreviation: - n/a
    - iso:country: `US`
    - iso:parent: - n/a
2. **Puerto Rico**
  - **option 1:** it's a valid ISO 3166-1 "country", and WOF think's it's an admin-0 `dependency` that is equal to `country`. 
    - wof:country: `PR` see https://github.com/whosonfirst-data/whosonfirst-data/issues/436 
    - wof:abbreviation: - n/a
    - iso:country: `PR`
    - iso:parent: - n/a
  - **option 2:** it's a valid ISO 3166-1 "country", and WOF think's it's an admin-0 `dependency` that is NOT equal to `country`. 
    - wof:country: `US`
    - wof:abbreviation: - n/a
    - iso:country: `PR`
    - iso:parent: - n/a
  - **option 3:** it's a valid ISO 3166-1 "country", and WOF think's it's an admin-0 `dependency` that is **NOT** equal to `country`, but we'll tell you which "country"
    - wof:country: `US`
    - wof:abbreviation: - `PR`
    - iso:country: `PR`
    - iso:parent: - n/a
  - **option 4:** it's a valid ISO 3166-1 "country" that's actually a dependency of US, and WOF think's it's an admin-0 `dependency` that is equal to `country` and look up the hierarchy if you really care.
    - wof:country: `PR`
    - wof:abbreviation: - n/a
    - iso:country: `PR`
    - iso:parent: - `US`
  - **option 5:** it's a valid ISO 3166-1 "country" that's actually a dependency of US, and WOF think's it's an admin-0 `dependency` that is NOT equal to `country`, and we'll do the lookup for you:
    - wof:country: `US`
    - wof:abbreviation: - `PR`
    - iso:country: `PR`
    - iso:parent: - `US`
3. **French Guyana**
  - valid ISO 3166-1 "country" (and oddly enough an 3166-2 "region"), and WOF think's it's an admin-1 `region` that is NOT equal to `country`. 
  - **option 1:** 
    - wof:country: `FR`
    - wof:abbreviation: - n/a
    - iso:country: `GF`
    - iso:parent: - n/a
  - **option 2:** n/a
  - **option 3:**
    - wof:country: `FR`
    - wof:abbreviation: - `GF`
    - iso:country: `GF`
    - iso:parent: - n/a
  - **option 4:** a variant of 3 but in ISO properties, which aren't really ours
    - wof:country: `FR`
    - wof:abbreviation: - n/a
    - iso:country: `GF`
    - iso:parent: - `FR` - most similar issue: https://github.com/whosonfirst-data/whosonfirst-data/issues/211
  - **option 5:** best/worst of all worlds
    - wof:country: `FR`
    - wof:abbreviation: - `GF`
    - iso:country: `GF`
    - iso:parent: - `FR`


Once we settle on what's what... please document the properties for ISO and WOF described in https://github.com/whosonfirst-data/whosonfirst-data/pull/823#issue-235393805 in the properties repo:

- https://github.com/whosonfirst/whosonfirst-properties/blob/master/properties/wof.md
- add iso.md

thisisaaronland:
I need to step away from the computer but I will look at all these words when I return.

nvkelso:
@dianashk @orangejulius @trescube @thisisaaronland @stepps00 Can you review https://github.com/whosonfirst-data/whosonfirst-data/pull/823#issuecomment-316213254 and confirm:

- okay to drop `iso:parent` property
- your use and assumed meaning of `wof:country` property
- your use and assumed meaning of `iso:country` property
- your use and assumed meaning of `wof:abbreviation` property

dianashk:
So many words! Thanks @nvkelso for taking the time to think through this in such detail.

I've looked through our WOF related code and here's all I could find.

1. When setting the abbreviation property of a WOF record we do the following:
    * if `country` placetype, use `wof:country_alpha3`
       * for countries we just grab the abbreviation from `wof:country_alpha3` to avoid doing the conversion from alpha2->alpha3, which is what we've committed to using in our abbreviations for countries
    * anything other than country, use `wof:abbreviation`
       * we are relying on the `wof:abbreviation` to be present for regions, macroregions, dependencies, and countries, where appropriate

2. When setting the default name property of a record
    * use `iso:country` to check if the record is in the US so we can treat US counties in a special way (i.e. set their names to `qs:a2_alt` where available)

3. In the `placeholder` service, when indexing all the alt-names for WOF records
   * if placetype is `country` or `dependency` 
      * we index `ne:iso_a2`, `ne:iso_a3`, `iso:country`, and `wof:country_alpha3`
      * we use either `wof:country_alpha3` or `ne:iso_a3` for default abbreviation
   * if placetype anything other than country, we use `wof:abbreviation` for default abbreviation
   * also use `iso:country` to check if the record is in the US, similar to 2 above

I believe that's all we do with `abbreviation`, `country`, `parent` properties at this time. We do not have any current plans to change what we're doing with these, unless WOF changes and we need to update our code to avoid errors.

thisisaaronland:
I don't have strong opinions about `iso:parent` or `wof:abbreviation`  although I'm not sure that either or both can't be addressed better elsewhere. My memory is that both were introduced separately as a way to generate labels and/or address the lack of consistent "label" properties across records?

In as much as we're still trying to distill what constitutes a minimum set of properties (think of the records that @dphiffer generates from the simplified /venue tool in BI) and whether or not we have "roles" for properties like we do for placetypes, I would definitely put `iso:parent` and `wof:abbreviation` in the "optional" bucket.

If they are what it takes to get the job done today (and yesterday) then that is our burden and this is the work. Ideally I would prefer to see a full suite of language-specific abbreviations that can be used to generate language-specific `label:` properties complete with qualifiers (preferred, variant, etc.) like we do for `name:`

As for `wof:country` and `iso:country` those two should always be present. The latter allows people to dip in to WOF using ISO codes which are already used extensively across a variety of endeavours. The former is the 2-letter representation of a record's `country_id` and necessary because ISO has some weird ideas about countries (it's their party so whatever...) and to facilitate determining which repo any given file _should_ be stored in, absent a pre-existing `wof:repo` property:

https://whosonfirst.mapzen.com/data#github-repos

To that end the `unlc:subdivision` stuff is incomplete but should not be dropped unless there is a suitable replacement. If the Statoids stuff can double as a global two-letter code for regions (+country code) then all the better.

https://github.com/whosonfirst-data/whosonfirst-data/issues/221

nvkelso:
Per IRL discussion as captured in https://github.com/whosonfirst/whosonfirst-properties/pull/35:

Suggest the following logic changes for Pelias (current logic will work, but some properties have been deprecated):

1. When setting the abbreviation property of a WOF record we do the following:
    * if `country` placetype, use `wof:country_alpha3`
       * for countries we just grab the abbreviation from `wof:country_alpha3` to avoid doing the conversion from alpha2->alpha3, which is what we've committed to using in our abbreviations for countries
    * anything other than country, use `wof:abbreviation`
       * we are relying on the `wof:abbreviation` to be present for regions, macro regions, dependencies, and countries, where appropriate

    * **UPDATE: WOF is going to switch to use `wof:shortcode` instead of `wof:country_alpha3` and `wof:abbreviation` for all placetypes. The earlier properties will remain, but won't be updated. We will also offer optional `label:{lang}_x_preferred_shortcode` for localization (e.g. use in placeholder service).**

2. When setting the default name property of a record
    * use `iso:country` to check if the record is in the US so we can treat US counties in a special way (i.e. set their names to `qs:a2_alt` where available)

    * **UPDATE: We made changes a while ago in WOF to store these names into `wof:label`, please switch over. There will be some additional options under `label:*` for this now, too, see the PR linked above for more discussion. See also: [102080351](https://whosonfirst.mapzen.com/spelunker/id/102080351/)**

3. In the `placeholder` service, when indexing all the alt-names for WOF records
   * if placetype is `country` or `dependency` 
      * we index `ne:iso_a2`, `ne:iso_a3`, `iso:country`, and `wof:country_alpha3`
      * we use either `wof:country_alpha3` or `ne:iso_a3` for default abbreviation
   * if placetype anything other than country, we use `wof:abbreviation` for default abbreviation
   * also use `iso:country` to check if the record is in the US, similar to 2 above

    * **UPDATE: See discussion above for `wof:shortcode` and `label:{lang}_x_preferred_abbreviation`**

@dianashk Can you please respond here with Pelias issues to adopt the new WOF-isms. Thanks!

stepps00:
Adding commit to swap `wof:abbreviation` with `wof:shortcode` ([wof properties](https://github.com/whosonfirst/whosonfirst-properties/tree/master/properties/wof)), then merging.

nvkelso:
Let's close this PR and continue the work in #1055 which is a cleaner PR.

PR #842, opened 2017-06-28
WIP: New min_zoom values

stepps00:
This is a WIP.


data/101/748/253/101748253.geojson
```diff
@@ -15,6 +15,7 @@
     "mps:latitude":38.920403,
     "mps:longitude":-6.33538,
     "mz:hierarchy_label":1,
+    "mz:min_zoom":"7.0",
```
nvkelso: Should be a float value (`7.0` not `"7.0"`).

stepps00:
Replaced.

Issue #843, opened 2017-06-29
Set English preferred names for non-English marinearea features


nvkelso:
This list should capture _most_ of the non-English `marinearea` features:

These will need to be translated into English (from the current `wof:name`).

We should also determine what the original language is and set that language's preferred name to the `wof:name` value.

scalerank | featurecla | name | note | min_zoom
--------- | ---------- | ---- | ---- | --------
2 | bay | Bahía de Campeche | Bay of Campeche | 4
3 | gulf | Golfo de California |  | 5
4 | gulf | Golfe du Lion |  | 5.7
4 | gulf | Río de la Plata |  | 5.7
5 | bay | Bahía Blanca |  | 6
5 | gulf | Golfo de Penas |  | 6
5 | bay | Bahía Grande |  | 6
5 | gulf | Golfo de Tehuantepec |  | 6
5 | gulf | Golfo San Mat’as |  | 6
5 | gulf | Golfo de Guayaquil |  | 6
5 | gulf | Golfo Corcovado |  | 6
5 | fjord | Boknafjorden |  | 6
5 | fjord | Storfjorden |  | 6
5 | fjord | Vestfjorden |  | 6
5 | fjord | Skagerrak |  | 6
7 | sound | Øresund | The Sound | 7.6
7 | bay | Mecklenburger Bucht |  | 7.6
7 | bay | Baía de Marajó | Internal Brazil Amazon River Waters | 7.6
7 | gulf | Golfo de Urabá |  | 7.6
7 | bay | Baía de São Marcos |  | 7.6
7 | lagoon | Waddenzee |  | 7.6
7 | fjord | Sognefjorden |  | 7.6
7 | fjord | Trondheimsfjorden |  | 7.6
8 | sound | Seno de Skyring | _untitled_36 | 7.7
8 | sound | Seno Otway | _untitled_37 | 7.7
8 | channel | Canal do Sul | Internal Brazil Amazon River Waters | 7.7
8 | lagoon | Ozero Mogotoyevo | _untitled_55 | 7.7
8 | bay | Guba Gusinaya | _untitled_56 | 7.7
8 | strait | Selat Dampier |  | 7.7
9 | channel | Canal Perigoso | Internal Brazil Amazon River Waters | 7.8
9 | channel | Canal do Norte | Internal Brazil Amazon River Waters | 7.8
9 | lagoon | Bras d'Or Lake | _untitled_46 | 7.8
[...]

PR #845, opened 2017-06-29
New wof:name value for marinearea records with no name

stepps00:
Related to https://github.com/whosonfirst-data/whosonfirst-data/issues/843.

This PR sets new `wof:name` values for `marinearea` records that currently have no `wof:name`. 

Should not require PIP work.


nvkelso:
Please also mark these all `"mz:is_funky": 1` as those names are fine for reverse geo but should never be displayed on a general purpose map.

stepps00:
https://github.com/whosonfirst-data/whosonfirst-data/pull/845/commits/f21eb72dbcfb5779a836fa699d91428d200a8739 takes care of https://github.com/whosonfirst-data/whosonfirst-data/pull/845#issuecomment-312068044.

nvkelso:
No PIPing, just name work.

PR #846, opened 2017-06-30
Zoom and NE prop updates - continent, country, pop places, regions

stepps00:
~This PR includes newly curated `mz:min_zoom` property values for 5800 of the 7350 records in the `ne_10m_populated_places.shp` file.~

~Features that did not match an existing record in WOF will be imported in a separate PR.~

To do:

- [x] Merge this PR with https://github.com/whosonfirst-data/whosonfirst-data/pull/851
- [x] Import all properties for **matched** features from the `populated_places` file
- [x] Once #660 is merged, update zoom levels for all regions. This should be done on a country-by-country basis with one-offs updated separately (like Moscow example)
- [ ] Take care of merge conflicts

PR #848, opened 2017-07-04
Min zoom updates - non-matched features

stepps00:
This PR includes 1,542 new records from the `ne_10m_populated_places.shp` file.

All records have been exportified, given required properties, and all `ne:*` properties (no special cases, as discussed). 

This follow-up to https://github.com/whosonfirst-data/whosonfirst-data/pull/846.


nvkelso:
Were these matched with any existing WOF features? (Using the meesoshapes import script.)

stepps00:
These features were the "leftovers" that did not match on the first round of `ne` concordance matching, and were imported as completely new records. 

I can run them through a modified version of the mesoshapes import script (they have different props and attributes to take into account) and adjust the PR accordingly.

nvkelso:
IRL: look at the quattroshapes point gazetteer import script instead (there are some that we spot checked that are already in WOF).

Also, `mz:min_zoom` needs to be additionally set off the `ne:min_zoom` value.

stepps00:
Closing in favor of https://github.com/whosonfirst-data/whosonfirst-data/pull/851

Issue #497, opened 2016-10-13
New Indian State: Telangana


tallytalwar:
As of 2nd June, 2014, Indian state [Andhra Pradesh](https://whosonfirst.mapzen.com/spelunker/id/85672141/#6/16.295/80.750) was divided and a new state "Telangana" was formed, previous this was a [locality](https://whosonfirst.mapzen.com/spelunker/id/890501051/#6/16.295/80.750).

Reference Docs: 
https://en.wikipedia.org/wiki/File:India_Telangana_locator_map.svg
https://lists.openstreetmap.org/pipermail/talk-in/2014-May/001963.html
http://www.openstreetmap.org/search?query=telangana#map=4/17.81/97.16
https://en.wikipedia.org/wiki/States_and_union_territories_of_India
https://github.com/geohacker/india/blob/master/state/india_telengana.geojson

stepps00:
[This commit](https://github.com/whosonfirst-data/whosonfirst-data/commit/be538aafe8e77e8be66e03172a3e812faa709a33#diff-6d6e9d064a6d75ef96249d885c9f51b9R17) updates all `region` records in India and includes the new `region` of [Telangana](https://github.com/whosonfirst-data/whosonfirst-data/blob/be538aafe8e77e8be66e03172a3e812faa709a33/data/110/880/597/9/1108805979.geojson). 

Keeping this issue open until the commit is merged to `master`.

stepps00:
Related: https://github.com/whosonfirst-data/whosonfirst-data/issues/636

stepps00:
We recently imported a new region record for Telengana (!!), but the hierarchy is incorrect.

Leaving this issue open until https://github.com/whosonfirst-data/whosonfirst-data/pull/861 is merged.

stepps00:
Telengana has been imported and the `wof:hierarchy` has been updated to include the correct parents (India, Asia). 

Link to record in GitHub: [Telengana](https://raw.githubusercontent.com/whosonfirst-data/whosonfirst-data/master/data/110/880/597/9/1108805979.geojson).

Closing this issue shortly.

PR #861, opened 2017-07-14
Fixed Telengana parenting issue

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/497.

A new record for the region of Telengana, India was imported into Who's On First through a recent meso import, but the hierarchy was incorrectly set. Telengana is not current;y parented by India, but rather Kashmir.

This PR updates the hierarchy of the Telengana record. The PIP tools should update any descendant records.


stepps00:
This issue actually gets fixed by the PR for #660:

https://raw.githubusercontent.com/whosonfirst-data/whosonfirst-data/177888be126d735627ca2aed23fc5b54ba223c28/data/110/880/597/9/1108805979.geojson

Suggest we don't merge this and instead use the PR for #660 to fix this issue.

nvkelso:
Leaving this assigned to @stepps00 as we think #660 will fix this once merge (and it almost is). Confirm that – if so close this PR. Else move to @thisisaaronland.

stepps00:
I've verified that the `wof:hierarchy` for this record has been updated correctly through the work done in #660. 

Closing this issue.

Issue #479, opened 2016-09-23
Copy `meso:hasc_id` into concordances list


nvkelso:
For all the Mesoshape imports, we didn't notice ahead of time that `"meso:hasc_id":"CD.KN.KN"` should really also go into the concordance list (as `hasc:id`:*).

Once those PRs are landed, crawl over the features and "make it so".

nvkelso:
Such as: 

- https://github.com/whosonfirst-data/whosonfirst-data/pull/467
- https://github.com/whosonfirst-data/whosonfirst-data/pull/468
- https://github.com/whosonfirst-data/whosonfirst-data/pull/469
- https://github.com/whosonfirst-data/whosonfirst-data/pull/470
- https://github.com/whosonfirst-data/whosonfirst-data/pull/471
& etc

nvkelso:
Related: https://github.com/whosonfirst-data/whosonfirst-data/issues/480

stepps00:
Related: #515

stepps00:
Looks like this was taken care of in: https://github.com/whosonfirst-data/whosonfirst-data/commit/02831853557f2734e0a49b7be70790a8fb43037b

We should leave this issue open until the mesoshapes imports settle, though, to make sure we catch all `hasc` property fields in all `county` and `region` features.

stepps00:
PR  #869 tracks this issue. Commit 02831853557f2734e0a49b7be70790a8fb43037b took care of #515.

PR #869, opened 2017-07-25
New hasc:id concordances for meso features

stepps00:
This PR fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/479 and will free up records for the upcoming Statoids import.

- New `hasc:id` concordance values for ~5k records. These values came directly from the `"meso:hasc_id"` property. Records with existing meso concordances were not edited. There were a few dropped/edited values, including:
  - Some records in Zimbabwe have a `"meso:hasc_id"` value beginning with `0ZW`, which was changed to `ZW`.
  - `NULL`, `""`, and `0` values were not included.
  - Occassionally, a `"meso:hasc_id"` value ended with a `\t` character. These were not included.


stepps00:
Once approved, this PR will be merged via the pipeline tool (no PIP work required).

PR #882, opened 2017-08-03
Updated unlc concordance properties

### LINKED ISSUE #641: Add unlc:subdivision to wof:concordances property

stepps00:
Fixes #641. Replaces #822.

For any feature that current has a `unlc:subdivision` property, a `unlc:id` concordance value was appended to the record's `wof:concordances` list.

As commented in the issue, the original `unlc:subdivision` property was kept in each record.

This PR will be used to test the PR import pipeline tool.

There are an additional 124 records updated in this PR compared to #822. These records were either skipped by the original callback script (likely) or updated with a new concordance value (not likely).


nvkelso:
This is ready for pipeline tool testing :)

PR #874, opened 2017-07-28
Geonames names import

stepps00:
WIP.

This PR includes new `name:*` properties from the Geonames alternatenames.txt file. I want to spot-check records before merge, but no additional work should be needed.

List of the records updated [here](https://github.com/whosonfirst-data/whosonfirst-data/files/1184461/gn_names_diff.txt).

PR #884, opened 2017-08-04
Second round of geonames names import

stepps00:
Using https://github.com/mapzen/whosonfirst-toolbox/blob/master/scripts/create_gn_id_list.py, a new csv file was created that contained every `gn:id` concordance value in WOF, and the record's `wof:id`. 

Using that list, the geonames import script was run a second time against the `alternamenames.tsv` file, which caught and added new `name:*` properties to an additional 2,410 records.

Follows https://github.com/whosonfirst-data/whosonfirst-data/pull/874.


stepps00:
To do: 

- [ ] Add names and `gn:id` concordance for `marinearea`, `disputed`, and `dependency` records that @ndcartography reviewed.

nvkelso:
So far with the first set of 2140 features everything is looking good. I see the right names being added, and even abbreviations :)

nvkelso:
Regarding https://github.com/whosonfirst-data/whosonfirst-data/pull/884#issuecomment-320320846, we talked IRL and @stepps00 will work with @ndcartography on a followup PR for those stragglers.

stepps00:
This PR is being processed through the Boundary Issue pipeline tool (woo!), here: https://whosonfirst.mapzen.com/boundaryissues/pipeline/234/

I'll open a new PR to track https://github.com/whosonfirst-data/whosonfirst-data/pull/884#issuecomment-320320846 shortly.

PR #890, opened 2017-08-10
WIP: Statoids import - admin1

stepps00:
WIP.

This PR is the first pass at the import of Statoids data for the "primary" division in Statoids.

To do:

- [x] Fix `wof:population_rank` property
- [ ] Import data for "leftovers"
- [ ] Ensure that duplicate `name:` properties were not addded to records (see: `name:eng_x_preferred` property in `data/101/751/697/101751697.geojson`)

Script used to import the data: https://github.com/mapzen/whosonfirst-toolbox/blob/master/scripts/statoids_import_admin1.py

PR #891, opened 2017-08-11
Statoids import - admin1

stepps00:
Replaces https://github.com/whosonfirst-data/whosonfirst-data/pull/890.

This PR uses the "primary divisions" (`admin1`) list from Statoids to update and add properties to the appropriate WOF records.

The script used to complete this work is here: https://github.com/mapzen/whosonfirst-toolbox/blob/master/scripts/statoids_import_admin1.py

Out of the 4,234 features (rows) in the Statoids file, 423 did not match to a record in WOF. These "leftover" features will be tracked in a separate issue. They did not match with an existing record for a variety of reasons (Statoids is out of date, WOF records need to be updated, etc.).

Eight records received a `wof:placetype_alt` property equal to `region`; these are cases where WOF has the record (like Singapore or Marshall Islands), but at the `admin0` level.

Property edits only; no PIP work required.


data/101/751/663/101751663.geojson
```diff
@@ -198,8 +211,11 @@
     "wof:concordances":{
         "dbp:id":"Pecs",
         "fct:id":"11683d74-8f76-11e1-848f-cfd5bf3ef515",
+        "fips:code":"HU15",
         "gn:id":3046526,
         "gp:id":813653,
+        "hasc:id":"HU.PS",
+        "iso:id":"PS",
```
nvkelso: I expect an **iso:id** concordance to include the ISO country code like `HU-PS` (using a - instead of a . concat char like HASC).

data/101/751/667/101751667.geojson
```diff
@@ -158,10 +174,14 @@
...
     ],
-    "wof:lastmodified":1501265151,
+    "wof:lastmodified":1502412660,
     "wof:name":"Kaposv\u00e1r",
     "wof:parent_id":85684845,
     "wof:placetype":"locality",
+    "wof:placetype_alt":"",
+    "wof:placetype_local":"city of county right",
+    "wof:population":66245,
+    "wof:population_rank":8,
```
nvkelso: Nothing wrong here, but noting that the `wof:population_rank` scale is optimized for Tokyo being the largest **locality**. This will work for **region** placetypes fairly well, but may break down for **country** (we could add a couple more ranks to cover that case).

nvkelso: Followup in: https://github.com/whosonfirst/whosonfirst-properties/issues/31

data/101/751/663/101751663.geojson
```diff
@@ -187,6 +187,19 @@
     "src:geom":"quattroshapes",
     "src:geom_alt":[],
     "src:lbl:centroid":"yerbashapes",
+    "src:population":"statoids",
+    "statoids:area-km":"163",
```
nvkelso: to: `area_km`

data/101/751/663/101751663.geojson
```diff
@@ -187,6 +187,19 @@
...
+    "statoids:area-mi":"63",
```
nvkelso: to: `statoids:area_mi`

data/101/751/663/101751663.geojson
```diff
@@ -187,6 +187,19 @@
...
+    "statoids:as-of-date":"2011-10-01",
```
nvkelso: to: `statoids:as_of_date`

data/101/751/663/101751663.geojson
nvkelso: You can construct this by concatenating:

`statoids:country + "-" + statoids:iso`

thisisaaronland:
Is iso:id really necessary?

nvkelso:
> Is iso:id really necessary?

@thisisaaronland Yes as it's the ISO 3166-2 code for the "region" level versus the ISO 3166-1 code for the "country".

stepps00:
The most recent commit (f2cb729) includes the following updates:

- `"statoids:as-of-date"` -> `"statoids:as_of_date"`
- `"statoids:area-mi` -> `"statoids:area_mi"`
- `"statoids:area-km"` -> `"statoids:area_km"`
- `"mz:is_current"` is now `1` for all updated records (we know they're current)

In reference to https://github.com/whosonfirst-data/whosonfirst-data/pull/891#issuecomment-321927154, I've kept the `iso:id` concordance, but updated it to include the country code.

PR #906, opened 2017-08-24
Statoids - admin2 import for countries with good/great coverage

stepps00:
This PR includes updates at the `admin2` level. The updates include new `hasc:id` cocordance values and new properties harvested and created from Statoids data. Records in the following countries are updated with new Statoids data in this PR:

- Albania
- Argentina
- Bolivia
- Burundi
- Central African Republic
- Chile
- Dominican Republic
- El Salvador
- Equatorial Guinea
- Eritrea
- Gabon
- Georgia
- Ghana
- Guinea Bissau
- Iran
- Ivory Coast
- Kenya
- Kyrgyzstan
- Laos
- Malaysia
- Mali
- Mauritania
- Mongolia
- Morocco
- Mozambique
- New Zealand
- Nicaragua
- Nigeria
- Panama
- Republic of Congo
- Saudi Arabia
- Sierra Leone
- Somalia
- Swaziland
- Syria
- Tunisia

A [follow-up issue](LINK) has been opened to track any "leftover" records in these countries that were not updated in this PR. About 330 records in these countries share `hasc:id` concordances, so the follow-up issue will untangle those records, add new `hasc:id` concordances, and import Statoids data.

(A variant of) this script was used for this work: https://github.com/mapzen/whosonfirst-toolbox/blob/master/scripts/statoids_import_admin1.py

PR #920, opened 2017-08-31
Statoids admin2 import for countries with "ok" and "bad" coverage

stepps00:
WIP.

This PR is a follow-up to #903 and #906. It includes new `statoids:*` properties, population properties, and additional updated harvested from the statoids import data. Updates also clean up `hasc:id` concordance coverage.

Once this PR is merged, there will be ~12 leftover countries with extremely low or no hasc coverage that will still need to be imported. Those countries will be tracked in a separate PR.


stepps00:
https://whosonfirst.mapzen.com/boundaryissues/pipeline/314/

PR #921, opened 2017-09-01
Statoids - admin2 import for countries with (almost) no hasc coverage

stepps00:
A follow up to #920 (merge this PR once #920 is merged). 

This PR imports Statoids data for `admin2` records in countries that have "terrible" hasc coverage in WOF. These countries include:

- Japan
- Turkey
- Ukraine
- Venezuela
- USA
- Mexico
- Madagascar

Additional work will need to be done to reconcile the two datasets once merged.


data/102/080/629/102080629.geojson
```diff
@@ -3,8 +3,9 @@
   "type": "Feature",
   "properties": {
     "edtf:cessation":"uuuu",
-    "edtf:inception":"uuuu",
+    "edtf:inception":"2010-04-01",
```
nvkelso: IRL discussion... This inception date seems like it's for a census rather than for this feature.

Suggest putting the date into `src:population:date` instead (and resetting `edtf:inception` to `uuuu`). As this may affect other Statoids PRs I'm fine with a linking a new general followup issue here but merging this PR as-is.

Issue #925, opened 2017-09-06
Remap wof:country_alpha3 to wof:shortcode


stepps00:
Map the `wof:country_alpha3` property value to the `wof:shortcode` property for country, dependency, and disputed records.

Don't remove the `wof:country_alpha3` property; just copy the property value into the new property.

Related: #924

PR #940, opened 2017-09-11
New wof:shortcode props

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/925.

This PR updated records that had a `wof:country_alpha3` property, using that value for a new `wof:shortcode` property. The `wof:country_alpha3` property was left intact.

This is the first set of `wof:shortcode` properties for WOF.


stepps00:
https://whosonfirst.mapzen.com/boundaryissues/pipeline/356/

Issue #947, opened 2017-09-18
County "Nong" is duplicated


anyadvisor:
```
1108416125 | uuuu | LA | LA | 1500012904 | 16.4021 | 106.5079 | 0 | Nong
1108416125 | uuuu | LA | LA | 1473897779 | 16.4021 | 106.5079 | 0 | Nong
```
Those records are identical (apart from date modified and filehash

nvkelso:
Those WOF ids at the beginning look the same to me. Where are you seeing the duplicates, can you link to a file tree in Github or S3 buckets or API, please?

- https://whosonfirst.mapzen.com/spelunker/id/1108416125/

anyadvisor:
https://github.com/whosonfirst-data/whosonfirst-data/blob/master/meta/wof-county-latest.csv

Lines 899 & 902 I think (maybe 903)

nvkelso:
I see what you mean. The IDs and paths look the same to me. @thisisaaronland any idea what's happening here?

anyadvisor:
Another change should be:
```
uuuu | 85632733 |   | 10.61616 | 47.36795 | 421173973 | uuuu | -99 | -99 | 1501212044 | 0 | 0 | 421173973 | Ceerigaabo | 1108757789 | locality | 1108805621 |   | -99
```
->
```
uuuu | 85632733 |   | 10.61616 | 47.36795 | 421173973 | uuuu | XS | XS | 1501212044 | 0 | 0 | 421173973 | Erigavo | 1108757789 | locality | 1108805621 |   | XS
```

Source: https://en.wikipedia.org/wiki/Erigavo
Source: https://www.google.co.uk/maps/place/9%C2%B043'32.4%22N+46%C2%B016'32.1%22E/@10.0928769,47.0987615,5.5z/data=!4m5!3m4!1s0x0:0x0!8m2!3d9.725679!4d46.275584

I've found a few errors here and there, should I post them in this thread?

thisisaaronland:
I will take a look at this shortly. It smells like a bug or a hiccup in the https://github.com/whosonfirst/go-whosonfirst-meta code but I'll need to poke around a bit before I can say anything for certain.

anyadvisor:
I don't know if this is the correct place to ask, or if there is a forum or something, but how come the missing country ID's aren't reverse engineered from the wof_country, and vice versa using the dependency/country tables?

And what is the difference between an ID of 0 and an ID of -1?

[...]

PR #968, opened 2017-10-05
Deprecated duplicate, bunk Nong record

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/947.

When #660 landed, the record `421174877.geojson` slipped into the `data/110/841/612/5/` filepath. This duplicate file should never have been added, so it was:

- Deprecated
- Given an `mz:is_current` of `0`

Since the filepath and id don't match, I am unable to use the exportify tool.. but I've verified the JSON is correct.. The actual `data/110/841/612/5/1108416125.geojson` is fine as is and does not need any updates.

File(s) in GitHub, for reference: https://github.com/whosonfirst-data/whosonfirst-data/tree/master/data/110/841/612/5


nvkelso:
Please also point it (superseded by, even though we're deprecating it correctly?) to the legit record.

stepps00:
https://github.com/whosonfirst-data/whosonfirst-data/pull/968#issuecomment-334520982 fixed in ef4a5cd.

nvkelso:
As long as @thisisaaronland thinks the paths are okay, then I'm fine here.

thisisaaronland:
Seems fine to me.

stepps00:
https://whosonfirst.mapzen.com/boundaryissues/pipeline/371/

Issue #931, opened 2017-09-07
Import population data  and concordances for municipalities in the Netherlands


stepps00:
The municipalities of The Hague, Utrecht, Amsterdam, and Rotterdam have already been updated via past neighbourhood update issues. For these areas (and their descendants), import the government-sourced population data being used in #837.

Include:
- `wof:population`
- `src:poplulation`
- `wof:population_rank`
- source property code

stepps00:
cc @ZBSingleton to hand over the population data csv.

stepps00:
Add Friesland.

Population data: [Netherlands_population_update.txt](https://github.com/whosonfirst-data/whosonfirst-data/files/1392134/Netherlands_population_update.txt)

PR #996, opened 2017-10-17
Added population and source properties to Netherlands records

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/931.

PR includes:

- New `wof:population` prop (from source data)
- New `src:population` prop
- New `wof:population_rank` prop
- Test of new exporter logic to handle `wof:placetype_alt` placetypes in `wof:hierarchy`

Once this is merged, all future Netherlands work will be done through the typical neighbourhood import process.


stepps00:
cc @zbsingleton - these are the final changes to the Netherlands records we discussed.

PR #1007, opened 2017-10-19
Reworked und_x_* names for thailand

### LINKED ISSUE #531: Fix-up Thailand names (post Mesoshapes import)
Following up from the Russia mesoshape import and PR https://github.com/whosonfirst-data/whosonfirst-data/pull/516 the new names should be "name:tha_x_variant":

* data/110/873/109/9/1108731099.geojson

```
"name:und_x_variant":[
 +        "\u0e2d\u0e32\u0e01\u0e32\u0e28\u0e2d\u0e33\u0e19\u0e27\u0e22"
 +    ],
```

stepps00:
Fixes #531.

Removes the `und_x_variant` name property in any Thailand meso-sourced record and places the value into the `tha_x_variant` name property. Checks for duplicates, too.


data/110/873/139/5/1108731395.geojson
```diff
@@ -34,7 +34,7 @@
         "Khok Si Suphan",
         "Khok Sri Supan"
     ],
-    "name:und_x_variant":[
+    "name:tha_x_variant":[
```
nvkelso: If there is no primary, then the variant should go into the primary slot.

stepps00:
Most recent commit moves any `name:tha_x_variant`  properties to the `name:tha_x_preferred` property if the record was missing `name:tha_x_preferred`.

This should be ready for merge (via pipeline).

Issue #803, opened 2017-05-25
Two valid neighbourhood records were incorrectly deprecated in Alexandria


stepps00:
Two records in Alexandria were in more than one import file for https://github.com/whosonfirst-data/whosonfirst-data/issues/388. This caused the records to be imported correctly, but incorrectly deprecated.

Copy the deprecated records into new records with the superseded dance. Also verify that this work cleans up any leftover work in https://github.com/whosonfirst-data/whosonfirst-data/issues/746 (specifically that `1108724979` has correct hierarchy information).

**Records:**
https://whosonfirst.mapzen.com/spelunker/id/85875841/#15/38.8149/-77.0429
https://whosonfirst.mapzen.com/spelunker/id/85875845/#15/38.8149/-77.0429

stepps00:
After review, it looks like the Alexandria cessation and deprecated files were not processed correctly. Alexandria should only have descendant neighbourhood records that have a `src:geom` of `mz`, but has descendants from `quattroshapes`, too.. the PR for this issue needs to include adjustments to all non-mz geoms, and the reprocessing of all descendants in Alexandria.

PR #1015, opened 2017-10-24
Reworked neighbourhood records in alexandria

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/803.

- Two deprecated neighbourhood records in Alexandria were given a `wof:superseded_by` value. These records already had a `mz:is_current` property equal to 0 and an `edtf:deprecated` date.
- Two new neighbourhood records were created. These records supersede the deprecated records and include:
  - New `reversegeo:*` properties
  - New `lbl:*` properties
  - Updated `src:geom` properties
  - `mz:is_current` of 1

This will require PIP work.


stepps00:
Holding off on import until https://github.com/whosonfirst-data/whosonfirst-data/issues/803#issuecomment-403645663 is taken care of.

nvkelso:
This generally looks good... but please find the old missing shapes to add into this PR so we have complete coverage for neighbourhoods & etc in this locality. It's pretty high visibility in the capital area of the USA.

stepps00:
I found the original import files - looks like we can import this once I update records from the `wof_neighbourhoods.geojson` file.

PR #1019, opened 2017-10-26
lbl:bbox property for Alaska and Russia

### LINKED ISSUE #1018: Set lbl bbox of Alaska, Russia to not wrap
When I search for Alaska in Mapzen search the bbox returned is world spanning. Instead we should clip it to -180 so a tiny bit of the Aleutian Islands would be cut off but the mainland would be front and center.

When I search for Russia the same thing is true, but clip to 180.

stepps00:
Fixes #1018.

Added a `lbl:bbox` property to the records for Alaska and Russia.

Screenshots of the new `lbl` bounding boxes: 

screen shot 2017-10-26 at 11 36 44 am

screen shot 2017-10-26 at 11 48 19 am


data/856/885/73/85688573.geojson
```diff
@@ -8,17 +8,19 @@
     "edtf:cessation":"uuuu",
     "edtf:inception":"2010-04-01",
     "geom:area":313.995134,
+    "geom:area_square_m":1712881940620.911621,
     "geom:bbox":"-179.231086,51.175092,179.859681,71.441059",
     "geom:latitude":63.742991,
     "geom:longitude":-151.593418,
     "iso:country":"US",
     "iso:subdivision":"US-AK",
+    "lbl:bbox":"-179.231086,51.175092,-129.979511,71.441059",
```
nvkelso: Minor quibble: `-179.98` (versus `-179.231086`) to match Russia would be more in keeping?

nvkelso:
Property only edits, can be put thru the pipeline.

stepps00:
https://whosonfirst.mapzen.com/boundaryissues/pipeline/476/

Issue #971, opened 2017-10-05
Supersede duplicate imported City of X localities into their original representation [USA]


nvkelso:
This is a followup from reviewing https://github.com/whosonfirst-data/whosonfirst-data/pull/824#issuecomment-334619360.

Supersede duplicate localities into their better representation. 
Where better = the first one that often has a polygon already.

Examples:

- `City of Fortuna` into **Fortuna** (California)
- `City of San Francisco` into **San Francisco** (California)
- `City of College Park` into **College Park** (Maryland)

This is _probably_ just a USA problem?

stepps00:
https://github.com/whosonfirst-data/whosonfirst-data/pull/824 has been merged - this is now a priority issue and will be worked on through mid-November.

stepps00:
If any of these records have an existing `edtf:cessation` date, it should be moved to the `edtf:deprecated`. Deprecated =/= cessation.

PR #1025, opened 2017-11-03
Update City of * records

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/971.

This PR takes care of all "City of *" locality point records in the US. The only "leftover" records are the following:

- [102526247](https://whosonfirst.mapzen.com/spelunker/id/102526247), City of Colorado Springs Airport, a valid campus.
- [420512551](https://whosonfirst.mapzen.com/spelunker/id/420512551), City of Temple Terrace (a valid, non-qs_pg locality)
- [85929465](https://whosonfirst.mapzen.com/spelunker/id/85929465), City of Creede (a valid, non-qs_pg locality)
- [85977105](https://whosonfirst.mapzen.com/spelunker/id/85977105), City of the Sun (a valid, non-qs_pg locality)


nvkelso:
Are there any `"County of"` or `"City and County of"` or similar to also think about?

stepps00:
I'll take a look at all `County of *` records once I commit the `City of *` changes.

stepps00:
Re https://github.com/whosonfirst-data/whosonfirst-data/pull/1025#issuecomment-341844478: 

Happy to report that there are only two "County of" / "City and County of" records ([1](https://whosonfirst.mapzen.com/spelunker/id/85684081/), [2](https://whosonfirst.mapzen.com/spelunker/id/85683751)), neither of which were imported from the Quattroshapes point gazetteer work.

stepps00:
@nvkelso - untagged this as "WIP". Ready for review, unless this PR should include:

- Another API run to catch the remaining deprecated records that were not superseded (~10% of them)
or
- Any records should be tagged with `"mz:is_current":1`

nvkelso:
@stepps00 Wither the 800 that didn't match a polygon? I suspect this is caused by two cases:

1. Locality point is too far to be "near" or perfectly "inside" the intended polygon, but there should have been a match.
2. We have 2 points, highly unlikely (more likely there are dup points in the qs_pg import, since the original qs import was polygons only).
3. It's just a legit point, but we should move the "City of..." name to variant, and reset default name to just the ... part.

nvkelso:
Noting here the the superseded records will be a mix of placetypes (mostly `locality`, some `localadmin`) based on the vagaries of Geonames hierarchy and our import logic, so that's okay. 

Also noting here that if the "City of..." record was imported as not-current then it's expected to not have a full hierarchy – but the thing it's being superseded into should have one already (no additional work needed).

nvkelso:
Generally the API PIP matching seems to be working as expected, so once my comments in https://github.com/whosonfirst-data/whosonfirst-data/pull/1025#issuecomment-342978373 are addressed 👍.

nvkelso:
Of the 800 unresolved, @stepps00 used a new config for Mapzen Places search (hat tip to Aaron) in https://github.com/whosonfirst-data/whosonfirst-data/pull/1025/commits/88da72c0024038a2a57a2de10905de70adc5c1fd and https://github.com/whosonfirst-data/whosonfirst-data/pull/1025/commits/1ebd18e24c5472983728f590e7962070c599da75. We think there are less than 200 left after that. Time to move on – let's merge this.

nvkelso:
Fix more merge conflicts, OMG.

nvkelso:
Since the changed records are just point geoms we can go thru the pipeline. Over to @stepps00.

stepps00:
https://whosonfirst.mapzen.com/boundaryissues/pipeline/518/

Issue #972, opened 2017-10-05
Deprecate imported neighbourhoods in our curated cities


nvkelso:
This is a followup from reviewing https://github.com/whosonfirst-data/whosonfirst-data/pull/824#issuecomment-334619360.

Don't bother figuring out what to superseded them into just deprecate them en mass by intersecting with the polygons of the curated neighbourhoods (e.g. the ones in #945).

- `Central Waterfront` (is a dup, but ignore)
- `West of Twin Peaks` (looks to be new, but deprecate anyhow)

stepps00:
https://github.com/whosonfirst-data/whosonfirst-data/pull/824 has been merged - this is now a priority issue and will be worked on through mid-November.

stepps00:
[This file](https://github.com/whosonfirst-data/whosonfirst-data/files/1455014/locality_ids.txt) contains the `wof:id` of any locality that has curated neighbourhood records. 

A PR for this issue will deprecate any erroneous/new descendant neighbourhood point records for the localities listed in that file.

PR #1034, opened 2017-11-17
Non-US duplicate locality points

### LINKED ISSUE #973: Supersede duplicate imported localities into their best representation
This is a followup from reviewing https://github.com/whosonfirst-data/whosonfirst-data/pull/824#issuecomment-334619360, and is different than https://github.com/whosonfirst-data/whosonfirst-data/issues/971 which is USA focused.

This issue to global review of "nearby" localities to be flagged and evaluated if nearby of the same name are actually dup's of each other. Determine the best and supersede the other into the best.

stepps00:
WIP. Will fix #973.

- Commit 72865e2 supersedes any record with a `wof:name` that includes "City of " into its best representation.

To do:

- [x] Review all new `qs_pg` sourced point records. There may be a few additional cases of these records having existing duplicate records. Supersede records in any case.

Related: #971 / #972


nvkelso:
Sure! Let's put this through the pipeline, please.

stepps00:
https://whosonfirst.mapzen.com/boundaryissues/pipeline/531/

PR #1052, opened 2017-12-04
New shortcodes and label abbreviations for regions in six countries

### LINKED ISSUE #840: Add abbreviations to region records in six countries
Add a `label:eng_x_abbreviation` property and `wof:shortcode` property to regions in:

- Australia
- Brazil
- India
- China
- South Africa 
- Russia

This is causing Pelias to fail when searching for addresses with region abbreviations in these countries (being tracked in https://github.com/pelias/pelias/issues/601).

stepps00:
Fixes #840.

Adds new `wof:shortcode` and `label:eng_x_abbreviation` properties to all regions in:

- Australia
- Brazil
- India
- China
- South Africa
- Russia

New values are two-character codes, pulled from other `hasc:id`, `ne:postal`, and `iso` properties.


stepps00:
Property-only edits, no PIP work required.

PR #1076, opened 2018-01-30
Use German symbols in locality name

### LINKED ISSUE #1069: Weird characters in German
Hi there, we've found that from time to time german names have wrong encoding that leads to problems with german symbols (ä, ö, ü). For example:
https://whosonfirst.mapzen.com/spelunker/id/102063415/

tadjik1:
Should fix #1069


stepps00:
Thanks for taking a look, @tadjik1!

stepps00:
@tadjik1 - the [most recent commit](https://github.com/whosonfirst-data/whosonfirst-data/pull/1076/commits/309a6b3a963a68f3c31a524819141cda1724e145) adds a few additional name variations for the record in question and processes it through our [export tools](https://github.com/whosonfirst/py-mapzen-whosonfirst-export) for import.

You may notice the unicode characters (ex "\u00fc"), but they're now stored in the German namespaces.

nvkelso:
@tadjik1 is this problem more widespread than just that one feature name?

(The PR otherwise looks fine to me.)

stepps00:
@nvkelso - this is a more widespread issue, but I think it is covered in #52 and #183. 

Closing this issue, will track other cases in those issues above (or a new issue).

Issue #1077, opened 2018-02-01
Greenfield Town is the wrong name for Greenfield, MA


simonw:
One of our users was confused by "Greenfield Town, MA" in our search results.

https://places.whosonfirst.org/id/85950383

This does seem like an odd name for https://en.m.wikipedia.org/wiki/Greenfield,_Massachusetts

simonw:
It looks like there is a local controversy related to this: http://www.recorder.com/Greenfield-residents-will-have-chance-to-weigh-in-before-city-town-vote-14410321

Nevertheless, for Who's On First I feel "Greenfield" is a more correct name than "Greenfield Town". In fact that article suggests that the correct name would be "The City known as the Town of Greenfield".

simonw:
Hah, I just found this deprecated record for "City of Greenfield Town" https://places.whosonfirst.org/id/1125365187/

simonw:
Recent news story from December 21st 2017: http://nepr.net/post/identity-crisis-no-more-greenfield-mass-finally-calls-itself-city

stepps00:
Fixed via: https://github.com/whosonfirst-data/whosonfirst-data/pull/1078

PR #1078, opened 2018-02-01
Rename "Greenfield Town" to "Greenfield"

simonw:
See https://github.com/whosonfirst-data/whosonfirst-data/issues/1077


simonw:
Bah, I don't want to propose changing the newline at the end of the file but I've failed to submit a pull request that doesn't mess with that.

stepps00:
Hey @simonw - thanks for the PR. I'll check this branch out to take care of a few machine-generated properties and add the other names to this record.

We have the option to add variant names in the `name:eng_x_variant` property, so we can retain all names like "Greenfield Town", "Town of Greenfield", etc.

stepps00:
[Most recent commit](https://github.com/whosonfirst-data/whosonfirst-data/pull/1078/commits/3850cc7710a135717585f7c63fbbac9ecbda7aaa) saves the variant name and runs the record through the exportify tool for import.

PR #1185, opened 2018-04-24
Import leftover Statoids data for the Netherlands

### LINKED ISSUE #929: Import Statoids data for the Netherlands
Once https://github.com/whosonfirst-data/whosonfirst-data/issues/837 is complete, import Statoids data for all `admin1` and `admin2` records in the Netherlands. 

Because we are importing new, recent population data direct from a government source for all records in the Netherlands, import Statoids population data as a top-level Statoids property. Use the more recent population data to build `wof:population`, `wof:population_rank`, and `src:population`.

stepps00:
Fixes #929.

Adds Statoids data to any records in the Netherlands that did not have Statoids data previously imported (`admin2`).  

Netherlands already has Statoids data: [link](https://spelunker.whosonfirst.org/id/85633337/)
Region records already have Statoids data: [link](https://spelunker.whosonfirst.org/id/85633337/descendants/?exclude=nullisland&placetype=region)

This should wrap up the Statoids import for the Netherlands.


stepps00:
Note: Because @ZBSingleton updated Netherlands admin records with newer population data than what is in Statoids, I'm preferring existing population data to Statoids.

stepps00:
Pinging @nvkelso - can merge this once approved, property edits only.

PR #1191, opened 2018-05-03
[do not merge] New county records for Sweden

### LINKED ISSUE #1123: Add municipality records for Sweden
WOF is missing the 290 [municipalities of Sweden](https://en.wikipedia.org/wiki/Municipalities_of_Sweden); these would fall into the `county` placetype.

See: https://en.wikipedia.org/wiki/Administrative_divisions_of_Sweden

stepps00:
Fixes #1123.
Source: https://github.com/whosonfirst/whosonfirst-sources/blob/master/sources/se-lant.json

Adds 290 new records for municipalities (_kommun_) of Sweden. Includes source properties, Mapshaper lat/long values and Swedish and English names. 

Requires PIP work.


stepps00:
Done through: https://github.com/whosonfirst-data/whosonfirst-data/pull/1227

Issue #1110, opened 2018-04-04
Update localadmin records in Sweden


stepps00:
WOF should have 2,523 `localadmin` records to represent the municipal subdivisions of Sweden. 

See: https://en.wikipedia.org/wiki/Administrative_divisions_of_Sweden

PR #1194, opened 2018-05-07
[do not merge] Updated localadmin records in Sweden

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/1110.

This PR includes:

- 1,639 (out of the existing 1,680) updated `localadmin` records in Sweden. Updates include:
  - new geometries, demoting the existing point geometries to alt-geometry records
  - new source properties
  -  new Swedish and English names
  - new mapshaper-sourced lat/long values
- 884 new `localadmin` records in Sweden
- 51 deprecated `localadmin` records in Sweden

To gain all possible join matches, `wof:id` values were appended to the source geometries through three separate joins:
- A join on the `name` + concatenated latitude value + concatenated longitude value 
- A join on the `name` property for leftovers (with duplicate checking)
- A spatial join (with name and duplicate checking)

All counts match source files and all possible matches should be included.

Source: https://github.com/whosonfirst/whosonfirst-sources/blob/master/sources/se-lant.json


stepps00:
Done through: https://github.com/whosonfirst-data/whosonfirst-data/pull/1227

Issue #1201, opened 2018-05-21
Correct wk prefix on many properties to wd


nvkelso:
Looks like when we landed data from Wikidata a while back we incorrectly used the `wk` prefix for Wikipedia instead of `wd` for Wikidata. These have very different licenses, so let's indicate the correct source 🤕!

Related: https://github.com/whosonfirst-data/whosonfirst-data/issues/827

NOTE: Once this is done, we'll need to update the sources and properties repos again, specifically re-running the `make all` script that calculates how a source is used. Wikipedia should only be `concordance`.

PR #1208, opened 2018-06-01
Swapped wk prefix for wd prefix

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/1201 and unblocks https://github.com/whosonfirst/whosonfirst-properties/issues/51.

Crawled the wof-data repo to catch any instance of a property containing the prefix of `wk:` (wikipedia) and changed that to `wd:` (wikidata).

This commit touches many, many files, so the best way to review changes is probably to poke around the `data` directory here: https://github.com/whosonfirst-data/whosonfirst-data/tree/stepps00/wk-wd-props/data

Can provide a list of updated ids if that is more useful. Property-only edits, no PIP work required.


nvkelso:
I spot checked a few of these, LGTM

thisisaaronland:
This commit doesn't make any sense, see notes here: https://github.com/whosonfirst/whosonfirst-properties/commit/99258a713f7ac6a1198d951f46199c089178f933#commitcomment-29547175

nvkelso:
Responded over there...

PR #1211, opened 2018-06-04
Updated Iraq record and alt-geometries

### LINKED ISSUE #1207: Investigate new country shape for Iraq
The country shape of Iraq does not include the northern regions, causing discrepencies in the hierarchies and country assignments of various descendant records. 

These region records are in Iraq, but not part of the country geometry of Iraq:

https://spelunker.whosonfirst.org/id/85672417/
https://spelunker.whosonfirst.org/id/85672475/
https://spelunker.whosonfirst.org/id/85672425/

screen shot 2018-06-01 at 9 59 37 am

We store several alt-geometries that include these areas. Suggest swapping the default geometry with an alt. See: https://github.com/whosonfirst-data/whosonfirst-data/tree/master/data/856/321/91/

stepps00:
Fixes #1207.

- Swapped the existing quattroshapes-sourced geometry for the meso geometry
- Stored the quattroshapes geometry as an alt-geometry
- Stored the meso geometry as a reversegeo alt-geometry
- Updated record's `lbl:latitude`, `lbl:longitude`, `mps:latitude`, and `mps:longitude` values to match the meso geometry's Mapshaper values
- Added a pointer property in the record to the reversegeo polygon.
- Removed the existing meso alt-geometry file (used as default geom now)


data/856/321/91/85632191.geojson
```diff
@@ -933,10 +933,11 @@
     "qs:level":"adm0",
     "qs:scale":200000,
     "qs:source":"US State Department, with Natural Earth mods",
-    "src:geom":"quattroshapes",
+    "reversegeo:geometry":"85632191-alt-meso-reversegeo.geojson",
```
nvkelso: Technically not necessary since the default geom is the same, but sure.

stepps00:
@nvkelso - this PR is ready for review.

The descendant records of the "missing" regions of Iraq (Kurdistan area) actually have Iraq in their hierarchy already. I've checked the hierarchies of these descendant records and they do not need to be updated.

Ideally, we would PIP descendant records regardless to catch any new descendants, but I've spot-checked and don't think that would be the case in this area of Iraq.

PR #1240, opened 2018-06-29
Region updates

### LINKED ISSUE #1127: Update Rose Atoll and Bajo Nuevo Bank
These regions should fall under the [United States of America](https://spelunker.whosonfirst.org/id/136253057/) empire record, not the [United States](https://spelunker.whosonfirst.org/id/85633793/) country record.

This is causing the United States to [parent 53 region records](https://spelunker.whosonfirst.org/id/85633793/descendants/?exclude=nullisland&placetype=region) instead of 51 (50 states + D.C.).

See:
- https://spelunker.whosonfirst.org/id/85668133/
- https://spelunker.whosonfirst.org/id/85681155/

stepps00:
Fixes #1127.

Updates two region records and one disputed record to be parented by the empire record of the United States of America, rather than the country record of the United States.


data/856/325/63/85632563.geojson
```diff
@@ -381,12 +382,12 @@
         },
         {
             "continent_id":102191575,
-            "country_id":85633793,
+            "country_id":136253057,
```
nvkelso: This is an empire, not a country.

PR #1259, opened 2018-07-17
Various: neighbourhood geometry swap

### LINKED ISSUE #1258: Use mz:tier_locality to set mz:hierarchy_label
As a follow-up to https://github.com/whosonfirst-data/whosonfirst-data/issues/149, use the `mz:tier_locality` property to set `mz:hierarchy_label` property.

If a neighbourhood record has the `mz:tier_locality` property set to `1`, `2`, or `3` (or a null value), set the `mz:hierarchy_label` property to `0`.

stepps00:
Fixes #1258.

This is a work in progress.

Follow-up commits are needed to:

- [x] reset any record that was already a point (this work should only occur on records with polygon geometries).
- [x] PIP records
- [x] rework `src:geom_alt` property

This PR swaps neighbourhood geometries for points and sets `mz:hierarchy_label` properties. Macrohood and microhood records were not affected.


nvkelso:
We're expecting many neighbourhood hierarchies to changes with this PR (limited to USA and primarily zetashapes with very low Flickr input photo counts), primarily because the old geometries were too big and their centroid created hierarchies that were bunk. We expect the new hierarchies to generally be better but still might have problems. One step at a time...

nvkelso:
This seems to generally be doing the right things...

But sometimes it's setting a duplicate alt_geom in the listing:

- https://github.com/whosonfirst-data/whosonfirst-data/blame/stepps00/hierarchy-labels/data/858/049/57/85804957.geojson

Clean that up and you should be good to go.

stepps00:
Re https://github.com/whosonfirst-data/whosonfirst-data/pull/1259#issuecomment-406742541:

This was taken care of in commit: https://github.com/whosonfirst-data/whosonfirst-data/pull/1259/commits/c9152dc6909ff1667d00b85af9124a3db12f2a43

I've also filed a general follow-up issue to track other cases outside of this PR in #1262.

I will circle back for another review this afternoon before merging.

stepps00:
:+1:

PR #1261, opened 2018-07-20
Fix currency and date of La Lengua

stepps00:
One off fix, La Lengua should be "current".


data/102/112/179/102112179.geojson
```diff
@@ -34,9 +33,6 @@
     "name:spa_x_preferred":[
         "Academia Salvadore\u00f1a de la Lengua"
     ],
-    "name:unk_x_variant":[
```
nvkelso: Please add this back in so we can crosswalk with the gold plated shit some people shill.

stepps00: fixed.

Issue #1260, opened 2018-07-17
Append Township to USA localadmin of US Census twp type


nvkelso:
For example, this place is actually `Ohio Township, Indiana`, which makes more sense than `Ohio, Indiana`. 

- https://spelunker.whosonfirst.org/id/404501485/

This is likely a subtask of the mega issue of:

- https://github.com/whosonfirst-data/whosonfirst-data/issues/538

PR #1269, opened 2018-07-26
Name swapping for townships

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/1260.

Used the list of ids in the issue to update `wof:name`, `name:eng_x_preferred` and `name:eng_x_variant` properties for selected township records.

PR includes:
- Updated `wof:name` that includes the original name + ' Township'
- Moved existing preferred name(s) to the variant namespace
- Updated the preferred name to be equal to the new `wof:name`


stepps00:
Comments addressed in: 959c9eeb30ce7. Merging.

PR #1280, opened 2018-08-07
Updates to China, India, Pakistan, etc...

### LINKED ISSUE #169: Jammu and Kashmir needs to be split in three separate records
856/721/37/85672137.geojson

stepps00:
WIP. Will fix #169.

- [ ] Update Pakistan, India, and China geometries
- [ ] Update appropriate region records
- [ ] Update appropriate disputed records
- [ ] PIP descendants


stepps00:
Most recent commit includes PIP work on records and descendants.. PR is ready for review.

/cc @nvkelso

data/110/869/218/5/1108692185.geojson
```diff
@@ -48,19 +49,21 @@
             "continent_id":102191569,
             "country_id":85632659,
             "county_id":1108692185,
+            "disputed_id":1159339495,
             "region_id":85675745
         },
         {
             "continent_id":102191569,
-            "country_id":85632659,
+            "country_id":85632469,
```
nvkelso: Why is the country of India coming thru here? Is it being picked up via PIPing via the disputed territory? It's not wrong per-say... but it's a little odd since India would claim it's part of Jammu and Kashmir not region `85675745` which is a Pakistan point of view.

stepps00: This is occurring because the `disputed` record of Gilgit-Baltistan is disputed between Pakistan and India:  https://spelunker.whosonfirst.org/id/1159339495/

PIP work trickles this disputed hierarchy to descendants..

nvkelso: @stepps00 Can you file a ticket on PIP logic to respect the new label hierarchy booleans when constructing the dual hierarchy, please? I thin there are enough flags in there to achieve this today as long as PIP logic looks at them?

stepps00: As in, when a `mz:hierarchy_label_(placetype)` property is present and = 0, don't update descendant records at that (placetype)..?

Issue #1266, opened 2018-07-24
Review potential duplicate records


stepps00:
[This file](https://github.com/whosonfirst-data/whosonfirst-data/files/2225395/duplicate_wof_records.txt) contains wof:ids of potentially duplicate records. 

Matches of only two ids should simply need supersedes/superseded_by work, while the matches with three or more ids will need some investigation.

Matches were only flagged if names and placetypes are identical.

stepps00:
Related issues:
- https://github.com/whosonfirst-data/whosonfirst-data/issues/1138
- https://github.com/whosonfirst-data/whosonfirst-data/issues/1098
- https://github.com/whosonfirst-data/whosonfirst-data/issues/1086
- https://github.com/whosonfirst-data/whosonfirst-data/issues/1082
- https://github.com/whosonfirst-data/whosonfirst-data/issues/888

stepps00:
The majority of these cases have been fixed, via: #1283. However, there were about 2k "duplicate" sets that have not been taken care of.

~1650 of these sets were deemed "not duplicates", as they are actually separate features (chomes in Japan, locality/localadmin in the US and CA, neighbourhoods in Korea/Japan); they should stay as-is.

~300 duplicate sets that still need to be reviewed and will likely require more work in a new PR. These are mostly exploded multi-part polygons. 

Keeping this issue open until a new PR is opened to address these leftovers.

Lists of ids: [set one](https://github.com/whosonfirst-data/whosonfirst-data/files/2319215/skip_coupled_pairs_w_two_polygons_in_US_JP_etc.txt), [set two](https://github.com/whosonfirst-data/whosonfirst-data/files/2319216/skip_four_plus_duplicates.txt), [set three](https://github.com/whosonfirst-data/whosonfirst-data/files/2319217/skip_japan_and_us_duplicates.txt), [set four](https://github.com/whosonfirst-data/whosonfirst-data/files/2319218/skip_triplepairs_two_or_more_polygons.txt).

Issue #1087, opened 2018-03-13
Trim name fields


missinglink:
Heya,

I noticed that this record has a trailing space in the name, would it be possible to programmatically trim spaces from names when they are added/processed?

```
{
"id": 102063261,
...
"name:eng_x_preferred": [
"München "
],
```

Tx!

stepps00:
@missinglink - definitely possible to catch other instances of these trailing spaces. I'll update this issue with any similar cases.

aaparmeggiani:
If my count is wright, in the same situation you have .. 
```
sqlite> select count(*) from names where name like '% ';
70117
```
with the top culprits .. 

```
sqlite> select id, count(*) as count from names where name like '% ' group by id order by count desc;
     id     | count
------------+-------
 1125892065 |   174
  101748691 |    67
  101819833 |    65
  101819723 |    64
 1125905623 |    64
 1125945577 |    64
 1125897693 |    64
 1126049711 |    64
 1126011967 |    64
 1125786193 |    63
```

aaparmeggiani:
Not only whitespaces, but I also found a couple of \n (s) which are being carried to the sqlite files and breaking csv exports .. 

```
gis=# select id, name, regexp_matches(name, '[\n]+' ) from spr;
     id     |             name             | regexp_matches
------------+------------------------------+----------------
   85801073 | Vivier d'Oie                +| {"            +
            |                              | "}
 1108943237 | Tosko Celo                  +| {"            +
            |                              | "}
 1108943249 | Zgornja Hrusica             +| {"            +
            |                              | "}
 1108943285 | Spodnja Zadobrova           +| {"            +
            |                              | "}
 1108943287 | Zgornja Zadobrova           +| {"            +
            |                              | "}
 1108943373 | Dobrava pri Crnucah         +| {"            +
            |                              | "}
 1108946249 |                             +| {"            +
            | Les Pentes croix-roussiennes | "}
 1108946301 |                             +| {"            +
            | Montchat                     | "}
 1108950881 |                             +| {"            +
            | Arnavutkoy Merkez            | "}
(9 rows)
```

Ex: Rel [#1108946301](https://raw.githubusercontent.com/whosonfirst-data/whosonfirst-data/master/data/110/894/630/1/1108946301.geojson)
```
    "wof:id":1108946301,
    "wof:lastmodified":1496506772,
    "wof:name":"\nMontchat",
    "wof:parent_id":101749431,
[...]

PR #1281, opened 2018-08-08
[wip] Supersede work for duplicate records

stepps00:
WIP. This will fix (most of) #1266.

This PR will include the following:
- Superseding one record into the other, bringing any qs, gn, concordances, or alt names over.

This work is only being done on duplicate pairs with one or two point geometries. This work is not being done on duplicate sets of three or more records, or duplicate sets that have two or more geometries that need to be merged together.

To do:
- [x] Tag things as current
- [x] Scrub name fields, saving some names in the variant namespace.


data/101/712/549/101712549.geojson
```diff
@@ -12,13 +12,17 @@
...
+    "gn:gn_pop":11169,
+    "gn:population":10899,
     "iso:country":"US",
     "lbl:latitude":41.433533,
     "lbl:longitude":-81.75184,
     "mps:latitude":41.433533,
     "mps:longitude":-81.75184,
     "mz:hierarchy_label":1,
-    "mz:is_current":-1,
+    "mz:is_current":1,
```
nvkelso: This is probably too aggressive across the entire PR, suggest leaving `mz:is_current` at -1.

data/101/712/549/101712549.geojson
```diff
@@ -207,13 +232,17 @@
     "wof:concordances":{
         "fct:id":"08d085c2-8f76-11e1-848f-cfd5bf3ef515",
         "fips:code":"3909246",
-        "gn:id":5110302,
+        "gn:id":5148346,
```
nvkelso: Please leave existing concordances alone, and add new ones from the dup record to `wof:concordances_alt` (the opposite of how you've done it here).

data/101/712/819/101712819.geojson
```diff
@@ -16,7 +16,7 @@
     "mps:latitude":40.200754,
     "mps:longitude":-84.800907,
     "mz:hierarchy_label":1,
-    "mz:is_current":-1,
+    "mz:is_current":1,
```
nvkelso: Unsetting the is_current commit should remove this feature from the diff, good!

data/101/725/033/101725033.geojson
```diff
@@ -139,10 +139,10 @@
         "\u0cb2\u0cbe\u0cb0\u0cc6\u0ca1\u0cca"
     ],
     "name:kor_x_preferred":[
-        "\ub77c\ub808\ub3c4"
+        "\ub7ec\ub808\uc774\ub3c4"
```
nvkelso: We should leave the preferred alone instead of swapping these.

nvkelso: (prefer value in the "master" record instead of the random records being superseded into it.)

data/101/803/089/101803089.geojson
```diff
@@ -178,7 +178,7 @@
         }
     ],
     "wof:id":101803089,
-    "wof:lastmodified":1527642767,
+    "wof:lastmodified":1533771393,
```
nvkelso: Why was this modified?

data/101/712/549/101712549.geojson
stepps00: 👍

data/101/712/549/101712549.geojson
stepps00: Noting that this occurred because I was (incorrectly) preferring the deprecated feature's concordances, instead of the current feature's concordances. Swapped logic in import script, will rerun...

data/101/713/017/101713017.geojson
```diff
@@ -66,6 +69,9 @@
     "name:srp_x_preferred":[
         "\u0410\u043d\u0442\u0438\u043e\u043a"
     ],
+    "name:srp_x_variant":[
+        "\u0410\u043d\u0442\u0438\u043e\u043a "
```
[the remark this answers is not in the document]
stepps00: I'll add logic to catch trailing spaces and remove names when they are exact (minus that trailing space).

Follow-up cases will occur through: https://github.com/whosonfirst-data/whosonfirst-data/issues/1087

stepps00:
Closing and replacing....

PR #1287, opened 2018-08-17
SFO campus geometry changes

thisisaaronland:
This PR:

* Swaps out the `woedb` geometry for the `uszcta` geometry for the 94128 postal code (which _is_ SFO)

* Adds a `-alt-woedb` alt file

* Updates the parent and hierarchy to be SF county because that is "reality" whatever the geography says; Updates `wof:controlled` accordingly

screen shot 2018-08-17 at 12 37 40


data/102/527/513/102527513.geojson
```diff
@@ -213,23 +213,27 @@
...
         {
             "campus_id":102527513,
             "continent_id":102191575,
             "country_id":85633793,
-            "county_id":102085387,
-            "locality_id":85922499,
-            "neighbourhood_id":85886431,
+            "county_id":102087579,
+            "locality_id":85922583,
+            "postalcode_id":554784711,
```
stepps00: :+1:

data/102/527/513/102527513.geojson
```diff
@@ -192,17 +194,15 @@
...
-        85688637
+        85922583,
+        102087579,
+        554784711
     ],
     "wof:breaches":[],
-    "wof:category":[
-        "airport"
-    ],
+    "wof:category":"airport",
```
stepps00: We've listed this prop as a `list` type, per: https://github.com/whosonfirst/whosonfirst-properties/blob/master/properties/wof/category.json

Seems like this edit is correct (category = string, categories = list), but the property json is incorrect. Thoughts?

stepps00:
Small comment about the `wof:category` property, but otherwise LGTM.

data/102/527/513/102527513.geojson
thisisaaronland: I will defer to you and/or @nvkelso on this - categories are opinionated, tags are a free-for-all (hence a list)

nvkelso:
Shouldn’t it be dual hierarchy? We do that for other multi-admin areas...

thisisaaronland:
@nvkelso - you are correct, and I only saw your comment after pressing `merge`. I will update things shortly.

thisisaaronland:
Fixed: https://github.com/whosonfirst-data/whosonfirst-data/pull/1289

PR #1288, opened 2018-08-17
Remove bunk variant name for Mauritius

### LINKED ISSUE #1286: Wrong `fra_x_variant` for Mauritius
`Île de France` is a wrong french variant for Mauritius [`id: 85632357`](https://spelunker.whosonfirst.org/id/85632357/). 

It's only an historical variant and it also conflicts with the real [`Île de France`](https://spelunker.whosonfirst.org/id/404227465/).

stepps00:
Fixes #1286.

Removes bunk variant name that is already stored in the `historical` namespace.

Property only, requires no PIP work.


nvkelso:
LGTM

Issue #1285, opened 2018-08-17
 Ensure src:geom property to austriaod records


stepps00:
https://spelunker.whosonfirst.org/id/1108838099/

Because sometimes the `src:geom` property is missing.

Issue #1275, opened 2018-08-01
Trace source of "unknown" geoms


stepps00:
[Sometimes](https://github.com/whosonfirst-data/whosonfirst-data/pull/1234/files/1e7e542b4112978bee3250e698999bb81cb47ac3#diff-d59bdefd461c52b928c37165edc2327eR69) WOF maintains records with a `src:geom` or `src:geom_alt` property value of "unknown".

Go through previous commits to figure out where these records were sourced from, crawl the repo, and update in a new PR.

stepps00:
This is still an issue, and something we should tackle sooner rather than later.. an example: https://spelunker.whosonfirst.org/id/890445001/

I'm guessing these came from either NaturalEarth or GeoNames, based on the vintage of the `wof:id` and source properties.

PR #1283, opened 2018-08-16
Supersedes work, etc for duplicates

stepps00:
Will fix (most of) #1266, replaces #1281.

#1266 includes a list of any duplicate records, found by matching records on a string of lat+long+placetype+name. This PR takes care of these issues, minus the following cases:

- Any duplicate matches in Japan
  - The "chome" neighbourhood records are unique and not duplicates.
- Duplicate pairs where both geometries are polygons
- Any sets of three or more records

Comments to be addressed in this PR:
- [x] ~10k features superseded into the remaining ~10k features
- [x] Set `mz:is_current` to `-1`
- [x] Leave existing concordances in `wof:concordances`
- [x] Remove any trailing spaces at the end of `name:` values
- [x] Keep preferred names in the preferred space, variants in the variant space
- [x] Exportify only superseded and superseding records
- [x] Add list of ids per commit for future QA

To do:
- [x] De-dup duplicate pairs where both geometries are polygons (2,300 pairs)
- [x] De-dup any sets of three or more duplicate records (900 sets)
    - [x] Do more analysis to figure out how many of the sets are single polygon with multiple points

PR #1290, opened 2018-08-20
Austria: add src:geom property and tag records as current

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/1285.

- Added `src:geom` property 
- Tagged as current

This clears up records in Austria, but there is more work to do in: #1275.


stepps00:
The majority of duplicate records that needed work in this PR were recently taken care of in #1283.

Upcoming commits will fix merge conflicts and take care of any leftovers...

stepps00:
7a0e0cb - ensures that all `austriaod`-sourced geometries are given a `src:geom` property with the correct value.
a1f4280 - ensures that all localadmin records with `Point` geometries are flagged as not current
a17970ecb6ac1 - superseded all localadmin records with a `Point` geometry into the appropriate current localadmin record and tagged them with `edtf:*` properties.

PR #1312, opened 2018-09-13
[wip] Fix records with more than one preferred name

### LINKED ISSUE #1311: Fix records with more than one preferred name
https://github.com/whosonfirst-data/whosonfirst-data/commit/320caff3b2d75935701f4e4313046416ac68dec2

Some records in this commit have more than on preferred name. Keep one name, store the rest as variant names.

stepps00:
Fixes #1311.

data/858/933/71/85893371.geojson was found to have more than one name stored in a preferred name space. Looking at the blame, there are other records from the same commit that were given more than one preferred name.

This PR keeps a single preferred name in the `name:(lang)_x_preferred` property simply moves additional names to the `name:(lang)_x_variant` property.

Note that some records (like the first one listed) didn't have their `wof:name` stored in a `name:` property to begin with. This PR does not add names, it only ensures a single value in the preferred property.


data/857/642/59/85764259.geojson
```diff
@@ -61,9 +61,9 @@
         "Presidente Roque Saenz Penya"
     ],
     "name:spa_x_preferred":[
-        "Departamento Presidente Roque S\u00e1enz Pe\u00f1a",
-        "Presidente Roque S\u00e1enz Pe\u00f1a"
+        "Departamento Presidente Roque S\u00e1enz Pe\u00f1a"
     ],
+    "name:spa_x_variant":[],
```
nvkelso: Empty Spanish variant?

data/857/968/61/85796861.geojson
```diff
@@ -47,9 +47,9 @@
         "Cornigliano"
     ],
     "name:ita_x_preferred":[
-        "Cornigliano Ligure",
-        "Cornigliano"
+        "Cornigliano Ligure"
     ],
+    "name:ita_x_variant":[],
```
nvkelso: Empty Italian variant (should be `Cornigliano`).

thisisaaronland:
It _is_ possible to have more than one preferred name. What's the rationale here?

data/857/968/61/85796861.geojson
stepps00: Good catch

stepps00:
Agreed that it is possible for records to have more than one preferred name.. this work is specific to the records updated in this commit: https://github.com/whosonfirst-data/whosonfirst-data/commit/320caff3b2d75935701f4e4313046416ac68dec2

These names should have been added in as variants, so Im swapping the values around for ~560 records.

stepps00:
Comments addressed in 11937d9, no name values were lost.

thisisaaronland:
👍

Issue #661, opened 2017-02-21
Upgrade neighbourhood records in Hyderabad, India


rmglennon:
This request came in through email support. Snips below:

Our state, Telangana (https://www.openstreetmap.org/relation/3250963),
and our workplace neighbourhood, Prashasan Nagar (https://www.openstreetmap.org/relation/6703400), are missing from the Who's on First gazetteer!

Can you forward this data to your team responsible for WOF? Or can you point us in the right direction where I can contribute?

_______________________________________
_(Edited by stepps00 on 02/21/2017)_

To-do for Hyderabad, India neighbourhood upgrades:

Status:
- [x] Propose **sources**
- [x] Establish **concordances** between WOF and source
- [x] **Upgrade / add** WOF shapes
- [x] **Add new or missing** features to WOF
- [x] **Deprecate / supersede** WOF features
- [x] **Upgrade** locality record's shape
- [x] **Draft review**
- [x] **Draft revisions** based on review comments
- [x] **Final review**
- [x] **Final revisions** based on review comments
- [x] **Import** into Who's On First
### Details

Sourcing:
- Permissive license?
- Add new src to Who's On First

Upgrade / add:
- `macrohood`
- `neighbourhood`
- `microhood`

Add new or missing features to WOF
- `macrohood`
- `neighbourhood`
- `microhood`

Deprecate / supersede:
- `macrohood`
- `neighbourhood`
- `microhood`

stepps00:
We have existing issue(s) open for new India data imports - [here](https://github.com/whosonfirst-data/whosonfirst-data/issues/636) and [here](https://github.com/whosonfirst-data/whosonfirst-data/issues/497). The new Telengana record will be imported into Who's On First once [this commit](https://github.com/whosonfirst-data/whosonfirst-data/pull/660/commits/be538aafe8e77e8be66e03172a3e812faa709a33) lands in `master`. 

We should update the neighbourhood records in Hyperbad, though, so I'll keep this issue open to track progress of the neighbourhood update work.

gopi-ar:
@stepps00 the major change of Telangana as a new state in India and also this neighbourhood (Prashasan Nagar) have been in OSM for quite a few years now, how come it's not updated to Pelias? We added "Unwired Labs" to OSM a couple of months ago and that shows up in Pelias just fine..

gopi-ar:
@rmglennon correction, it's "Hyderabad" though "Hyper-bad" sounds cooler. :-)

[...]

PR #1330, opened 2018-10-01
India: Update Telangana and descendants

stepps00:
Related to https://github.com/whosonfirst-data/whosonfirst-data/issues/661.

- Changes the name of **Telengana** to **Telangana**. This name was misspelled upon import, but the name properties are all correct, so only the `wof:name` changes.
- Added `reversegeo:*` properties to the [Hyderabad](https://spelunker.whosonfirst.org/id/102030059/) record so it correctly falls within the Hyderabad `county` record
- PIPs Telangana to update hierarchies of all descendant records (removing incorrect dual hierarchies). 

Keeping #661 open, since the original issue is to track neighbourhood updates in Hyderabad.


nvkelso:
Features inside Hyderabad should still receive dual hierarchies, though, as it is the capital of the Indian state of Telangana and de jure capital of Andhra Pradesh, right?

stepps00:
Yeah, that makes sense.

To do that, I'll need to:
- find any of the updated records in this PR with Telangana in their `wof:hierarchy`
- manually set dual hierarchies
- set `wof:controlled` properties
- update `wof:parent_id` to be `-4`

stepps00:
https://github.com/whosonfirst-data/whosonfirst-data/pull/1330#issuecomment-426409883 was taken care of in de05e50 and f5516d3, but I'd like to QA and set coterminous properties (locality and county of Hyderabad) before merging.

stepps00:
QAd the changes that things look good. 

I could set swap the existing geometry of the [locality of Hyderabad](https://spelunker.whosonfirst.org/id/1108696315/) with the smaller [county of Hyderabad](https://spelunker.whosonfirst.org/id/102030059/) geometry, but that would not cause additional hierarchy changes (they're already manually set). Plus, I'm not sure if I prefer one of those geometries over the other.

Thoughts @nvkelso?

nvkelso:
The two Hyderabad records both look wrong in different ways to me. Suggest no change for now.

PR #1348, opened 2018-10-23
Airports 2

thisisaaronland:
* update missing coordinate data; dates; country codes; historical names for: BHM,IDL,JFK,YYJ,INN and others
* update wof:name to be (default) english for a number of airports; move old wof:name to LANG_x_preferred


thisisaaronland:
Ugh... just noticed these all have `misc:` prefixes instead of `name:`. Go me!

PR #1353, opened 2018-10-24
WIP: GeoNames Import, rd 2

stepps00:
Round two of new GeoNames locality records.


data/124/272/007/5/1242720075.geojson
```diff
@@ -0,0 +1,78 @@
...
+            "country_id":85632465,
+            "county_id":1092049679,
+            "locality_id":1242720075,
+            "region_id":85675579
+        }
+    ],
+    "wof:id":1242720075,
+    "wof:lastmodified":1536857350,
+    "wof:name":"Ghaiyabari",
+    "wof:parent_id":1092049679,
```
nvkelso: Looks like the dual hierarchy has same county, but different regions. So do we expect the parent_id to be "complex" or an int for the county id?

nvkelso:
Besides my question about dual hierarchy, this looks same as last batch and generally fine to merge.

stepps00:
re https://github.com/whosonfirst-data/whosonfirst-data/pull/1353#discussion_r230242731: If the same parent id is present in all hierarchies, it is used to populate `wof:parent_id`, so this is expected.. this would only be "complex" if the parent ids (in this case county, or localadmin if it were in the hierarchy) were different values.

nvkelso:
Sounds good then.

PR #1372, opened 2018-11-02
Update ACT record

### LINKED ISSUE #1349: Australian Capital Territory abbreviation is not `CT`
Pelias users have reported that `CT` is not a valid or expected abbreviation for the Australian Capital Territory.

Currently, the WOF [ACT record](https://github.com/whosonfirst-data/whosonfirst-data/blob/master/data/856/814/93/85681493.geojson) has `CT` for `label:eng_x_abbreviation` and `wof:shortcode`.

Let me know if this is intended, or if you have information suggesting `CT` is ok to use in this case.

If it is correct in some cases, we'll look into using `abrv:eng_x_preferred` which is set to `ACT` for that record.

stepps00:
Fixes #1349.

Updates the ACT record with a new shortcode and label value, `CT` -> `ACT`.


orangejulius:
awesome, thanks for this!

PR #1374, opened 2018-11-02
Update Hinoba-an record in PH

### LINKED ISSUE #1365: Phillipine locality of Hinoba-an has an out of date name
The Philippine locality of [Hinoba-an](https://en.wikipedia.org/wiki/Hinoba-an) was referred to as `Asia` until a [1959 declaration](http://www.chanrobles.com/republicacts/republicactno2154.html) changed its name.

In WOF (and [Geonames](http://www.geonames.org/1730097/asia.html), although I am about to change it), it is still known as Asia: https://spelunker.whosonfirst.org/id/890416267/

For obvious reasons, this record can cause problems when searching for the continent, or vice versa.

Also note that the Wikipedia concordance for this locality is currently the Wiki page of the continent.

stepps00:
Fixes #1365.

Swaps the historical name of "Asia" into the historical name space, and sets the current name of "Hinoba-an" as the `wof:name` and `name:eng_x_preferred` value. Also removes the bunk `wk:page` concordance.


data/890/416/267/890416267.geojson
```diff
@@ -15,6 +15,12 @@
...
     "mz:hierarchy_label":1,
     "mz:is_current":1,
+    "name:eng_x_historical":[
+        "Asia"
+    ],
+    "name:eng_x_preferred":[
+        "Hinoba-an"
+    ],
     "name:fil_x_preferred":[
         "Asia"
```
nvkelso: Probably also update the `fil` name...

stepps00: adding in upcoming commit...

PR #1380, opened 2018-11-09
Philippines: Fixes to the Metro Manila and NCR records

### LINKED ISSUE #1378: Manilla
Heya,

The capital city of The Philippines "Manilla" (421168919) has been superseded by "Metro Manilla" (1159397249).

The new record includes the prefix "Metro" which is problematic for exact-matching search queries.

Could we please either:
- Add additional name/names of simply "Manilla" without the prefix.
- Update the existing names to remove the "Metro" part from the names.

Is there a WOF policy regarding these sort of things?
Eg. Should 'Kansas' be called 'Kansas State' and 'Berlin' as 'Stadt Berlin'?

It seems cleaner to omit the local name for the placetype and instead use `wof:placetype_local` to say this is a `metro`?

stepps00:
Fixes #1378.

Changes the name of "Metro Manila" to "Manila" and stores the existing name in the English variant property. Also takes the name translations and moves them to the National Capital Region record.


nvkelso:
Please keep the obvious `Manila` name translations on the locality (stripping the `metro` part).

PR #1384, opened 2018-11-13
Ukraine: Updated region records

### LINKED ISSUE #1371: Verify wof:name values in Ukraine
Verify that the `wof:name` values of Ukraine descendants are correct. Several `region` records have mismatched `wof:name` values and `name:*` properties, rework the name values when necessary.

See:
- https://spelunker.whosonfirst.org/id/85688877/
- https://spelunker.whosonfirst.org/id/85688819/

stepps00:
Fixes #1371 

This PR updates region records' (and two autonomous city records) properties, including:

- [ ] Verifying `wof:name` and `name:eng_x_*` property values are correct
- [ ] Updating `wof:placetype_local` values
- [ ] Swapping `statoids:` (and associated `wof:`) properties around (Luhansk and Kiev City were incorrectly swapped, and Sevastapol was added)

No geometry edits were made, no PIP work required.


data/856/888/01/85688801.geojson
```diff
@@ -80,11 +80,14 @@
         "Oblast Kiev"
     ],
     "name:eng_x_preferred":[
-        "Kyiv City"
+        "Kiev"
```
nvkelso: This should probably be something different than Kiev (since there's a city at the same level). So either keep this as `Kiev` and make the default English name of the actual city `Kiev City`, or rename this one to `Kiev Oblast`.

data/856/888/01/85688801.geojson
```diff
@@ -80,11 +80,14 @@
...
     ],
     "name:eng_x_variant":[
         "Kyivs'ka",
-        "Kyiv Oblast"
+        "Kyiv Oblast",
+        "Kyiv City",
```
nvkelso: This one isn't the city, so remove this line for the oblast.

PR #1403, opened 2018-11-26
China: Updated admin1 (and SAR records)

### LINKED ISSUE #1402: Update China admin1 names
Similar to #1397, update names and placetype values for admin1 records in China.

Also note that iso codes for China provinces changed in 2017 from the alpha-numeric codes that are currently in WOF.

https://www.iso.org/obp/ui/#iso:code:3166:CN

stepps00:
Fixes #1402.

- Updated `wof:name` when needed. Examples:
    - "Macao S.A.R" -> "Macao S.A.R."
    - "Ningxia" -> "Ningxia Hui"
- Stored `wof:name` value in `name:eng_x_preferred` property
- Updated and sanitized `name:eng_x_variant`, `name:zho_x_preferred`, and `name:zho_x_variant`property values
- Updated iso concordance values
- Added preferred placetype labels (and removed placetype_local)

See: 
  - https://en.wikipedia.org/wiki/Provinces_of_China
  - http://statoids.com/ucn.html
  - https://www.iso.org/obp/ui/#iso:code:3166:CN


stepps00:
Note that there may be additional commits to clean up Tibet/Xizang names.

data/856/324/03/85632403.geojson
```diff
@@ -15,6 +15,12 @@
     "itu:country_code":[
         "886"
     ],
+    "label:eng_x_preferred_placetype":[
+        "province"
```
nvkelso: This seems inconsistent (and problematic) considering the WOF placetype.

data/856/324/03/85632403.geojson
```diff
@@ -904,13 +911,13 @@
         "Tiong-ho\u00e2 B\u00een-kok"
     ],
     "name:zho_x_preferred":[
-        "\u4e2d\u83ef\u6c11\u570b"
+        "\u53f0\u6e7e"
```
nvkelso: Please ensure we have both traditional and simplified Chinese support here...

data/856/321/61/85632161.geojson
```diff
@@ -87,15 +93,16 @@
         "\u039c\u03b1\u03ba\u03ac\u03bf"
     ],
     "name:eng_x_preferred":[
-        "Macau"
+        "Macau S.A.R."
```
nvkelso: This makes it consistent with Hong Kong, but what a mouth full...

https://spelunker.whosonfirst.org/id/85632483/

nvkelso:
Please also have a look at Tibet and Xinjiang.

data/856/324/03/85632403.geojson
stepps00: Fair point, same would go for updates to Macao SAR and Hong Kong SAR..

data/856/324/03/85632403.geojson
stepps00: We're good here. Anytime I swap out the preferred name with a new value, I make sure the value is stored in the variant space.. in this case, the "\u4e2d\u83ef\u6c11\u570b" is in the `"name:zho_x_variant":` list (below).

data/856/321/61/85632161.geojson
stepps00: Noting here that I'll need to verify this issue does not affect Macau, too: https://github.com/whosonfirst-data/whosonfirst-data/pull/1403#discussion_r236519135

stepps00:
I'll also look into adding longname support.

PR #1420, opened 2018-12-03
Various: Update No Data and NULL names

### LINKED ISSUE #1411: Fix "No data" name
https://spelunker.whosonfirst.org/id/1108698631/

This county record has a `wof:name` of "No data". There may be a handful of other cases, so using the metafiles, check all county records' (likely just meso-sourced records) `name:` and `wof:name` properties and fix.

stepps00:
Fixes #1411.

Updates `wof:name` and `name:` properties for nine records that have either have "No Data" or "NULL" values. In one case, with 85672937 and 1108785435, I also flagged these records as coterminous.

I also flagged 1108698631 as not current, as it is not an administrative record, but a record for a Lake in the Philippines. This record could also stay current, with a land_use_aoi flag, but I'd prefer to get the "current" counts correct for admin2 in PH.


data/110/869/863/1/1108698631.geojson
```diff
@@ -2,10 +2,10 @@
   "id": 1108698631,
   "type": "Feature",
   "properties": {
-    "edtf:cessation":"uuuu",
+    "edtf:cessation":"2018-12-03",
```
nvkelso: As long as your adding this feature's area to the surrounding polygon, sure. But prove that here.

image

stepps00: The boundaries of each county surrounding this feature are actually clipped at the lake, so this county geometry does not fall within any administrative division, other than its parent region. 

It would be incorrect to split this geometry or dissolve it into a neighboring county, but we could keep it and flag it with `mz:is_landuse_aoi` = 1 if we want continuous county coverage in PH.

Thoughts @nvkelso ?

stepps00: https://en.wikipedia.org/wiki/Taraka,_Lanao_del_Sur
https://en.wikipedia.org/wiki/Tugaya,_Lanao_del_Sur
https://en.wikipedia.org/wiki/Lanao_del_Sur

nvkelso: We only use `is_landuse_aoi` for neighbourhoods generally. I'm fine using it here... but let's also use the `statistical_gore` flag = 1 since that's more common for these types of coarse admin geographies.

stepps00: I'll add a commit for these flags, as well as fixing merge conflicts then merge.

PR #1438, opened 2018-12-18
France: Locality fixes to Marseille, France...

stepps00:
One-off fix to correct a duplicate locality record for Marseille, France.

Requires PIP work.


stepps00:
4c73cff includes PIP work for Marseille, France & descendants.

/cc @nvkelso

PR #1469, opened 2019-01-24
Saudi Arabia: Updated county, locality, and neighbourhood records

### LINKED ISSUE #1468: Add polygon geometries to locality records in Saudi Arabia
Who's On First has [several thousand](https://spelunker.whosonfirst.org/id/85632253/descendants/?exclude=nullisland&placetype=locality) locality records in Saudi Arabia, but each record currently has a Point geometry. For the largest cities in this country, create polygon geometries by hand.

Also take care of any duplicate records in these areas, adjust parent geometries as needed, and demote locality points that fall within these new locality polygons to neighbourhoods. 

Start here:
- https://en.wikipedia.org/wiki/Template:Largest_cities_of_Saudi_Arabia

stepps00:
Fixes #1468.

This PR adds the following:
- New polygon geometries for the twenty largest localities (by population) in Saudi Arabia, as well as four smaller localities surrounding Dammam.
- Adjusts county geometries on the coastline, specifically around Dammam and Jeddah to cover all land areas and docks/ports.
- Supersedes any duplicate locality points, bringing over properties/concordances/names, when possible.
- Adjusts label centroids to fall over the urban core in cases where a new locality polygon has been created.
- Moves any locality point to a new neighbourhood record, if the locality point falls within a new locality polygon
- Exportified and validated all properties and geometries in any changes records

To do:
- [x] PIP records


nvkelso:
Please add in the next biggest 10 from that Wikipedia article, too.

stepps00:
https://github.com/whosonfirst-data/whosonfirst-data/pull/1469#issuecomment-457716561 addressed in 0548667. This commit gave new geometries to three region records and the country record for Saudi Arabia.

Upcoming commit will update hierarchies for all records in Saudi Arabia, then I'll merge.

Issue #1433, opened 2018-12-12
Llívia should be included in the geometry of Spain, removed from France


stepps00:
The geometry of this localadmin (exclave) should be added to the geometry of Spain and removed from the geometry of France.. https://spelunker.whosonfirst.org/id/404334755/

https://en.wikipedia.org/wiki/Ll%C3%ADvia

PR #1478, opened 2019-02-08
FR: Update geometry of France with cut-out for Llivia

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/1433.

After review, no changes were needed to Spain, or any descendants in Spain (macroregion, region, county, localadmin, or locality). Rather, the geometry for the country of France needed the "Llivia" geometry cut-out of it. 

Since the existing geometry of France already has a `src:geom` of `whosonfirst` (and a correct `src:geom_via` property), there is no need to update the src properies or store the bunk geometry as an alt. 

I've also reviewed hierarchies of descendants in the area and they all check out, no PIP work needed. This should be as simple as importing the new geometry. The record's properties and geometry validate.


stepps00:
See rich diff for geometry change:

screen shot 2019-02-08 at 3 20 33 pm

PR #1483, opened 2019-02-13
AUS: Updated admin records

### LINKED ISSUE #218: Investigate official Australian boundary data

stepps00:
This does not completely resolve #218, but is related.

This PR will contain:

- New `macrocounty` records in Australia, which represent the "significant urban areas"
- Updates to the `county` records, flagging them with a `mz:hierarchy_label` of 0
- Flagging any `locality` and `localadmin` records as coterminous, if they are in fact coterminous
- Deprecating any locality point (GeoNames or QS) if it falls within an existing locality polygon'
- Updates neighbourhoods, deprecating records that share names with their parent locality or, if names are unique, moving the neighbourhood to a point geometry

This PR will not update postalcode records or bring in new Australia data (aside from the SUA); that work will take place in a follow-up branch.

To do:
  - [x] Curate lbl centroids for new macrocounties
  - [x] https://github.com/whosonfirst-data/whosonfirst-data/pull/1483#issuecomment-463853226
  - [x] Update other placetype geometries (?)
  - [x] PIP work to update hierarchies
  - [x] Validate properties and geometries
  - [ ] Catalog follow-up work (see initial comment)


stepps00:
###
480b0a8cdafb0550f54807fd57b86d8b013be57d:
  - Adds new `macrocounty` records, representing the PSMA's SUA features

8864a805eda4a3eab0b6fa91f7c4f526aee7eaa3:
  - Flags all county records with a `mz:hierarchy_label` of `0`

91a6f5b580cae10d2ed092078e9ad4dc4e90029a:
  - Tags all coterminous localities and localadmin records

1795030cb8c40ec707842d4a29cda50be62ac736:
  - Deprecates and locality point that falls within a locality polygon
  - Does not swap them into a new placetype -- most of these cases can not be verified as actual neighbourhoods (or campuses, etc)

f9283552fe08ef77af8998a4705b31e5f8359e67:
  - If a neighbourhood record's `wof:name` is contained in the `wof:name` of it's parent locality, deprecates the neighbourhood and supersedes it into the locality.
  - This catches exact matches like a locality "Avalon" parenting a neighbourhood of "Avalon" and catches slight variations, like a locality of "The Rocks (Sydney)" parenting a neighbourhood of "The Rocks"

To do:
  - [ ] Neighbourhood work for records that do not match names with parent locality
  - [ ] PIP work to update hierarchies
  - [ ] Validate properties and geometries
  - [ ] Catalog follow-up work (see initial comment)

data/857/687/19/85768719.geojson
```diff
@@ -57,7 +57,8 @@
     "qs:woe_ver":"7.10.0",
     "src:geom":"quattroshapes",
     "src:geom_alt":[
-        "quattroshapes_pg"
+        "quattroshapes_pg",
+        "quattroshapes"
```
nvkelso: Slightly confusing we have a default geom of quattroshapes and an alt geom from the same provider?

nvkelso:
For the new `macrocounty` records, please also copy over PSMA's SUA feature concordance values?

data/857/687/19/85768719.geojson
stepps00: Ah, this is actually a mistake. The `src:geom` should be `whosonfirst` here. I'll update these records' values in a new commit.

nvkelso:
(We'll look at these changes on-screen together Friday.)

nvkelso:
We looked at this in person, and my final comments were addressed in https://github.com/whosonfirst-data/whosonfirst-data/pull/1483/commits/b06f483d4e347763a189df7a025cad19ab361e10.

LGTM!

stepps00:
I've verified that all geojson files validate, that all geometries are valid, and that all name properties validate. Good to merge.

stepps00:
Follow up: https://github.com/whosonfirst-data/whosonfirst-data/issues/1495

PR #1486, opened 2019-02-18
Adjust Haan polygon based on OSM Data

schorba:
https://nominatim.openstreetmap.org/details.php?place_id=198682599


thisisaaronland:
OSM data can not be imported in to WOF because of license incompatibilities (ODbL for OSM and CDLA 1.0 for WOF). Is there another openly license data source for this information?

Issue #1178, opened 2018-04-19
Senegal: region updates


stepps00:
WOF has 17 region records in Senegal, but should have 14. Investigate and propose a change.

https://spelunker.whosonfirst.org/id/85632365/descendants/?exclude=nullisland&placetype=region

See:
- https://en.wikipedia.org/wiki/Regions_of_Senegal

stepps00:
Also, QA concordance values, names, and any other properties that seem inconsistent.

PR #1504, opened 2019-03-01
Senegal: Update region records

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/1178

Updates all region records in Senegal. Updates include:

- Deprecating three region records with point geometries, superseding them into their appropriate/duplicate region record with a polygon record. 
- Moves over any properties and translations from the superseded record to the superseding record
- Updated `wof:name` and `name:eng_*` properties for current region records

I can add `label:` properties and make other adjustments, but I'd prefer a bulk issue to add `label:` properties to all admin1 records as a follow-up.

Property edits only, will not require PIP work. Superseded records have `Point` geometries and do not parent any features.


stepps00:
To do: 

- [ ] Validate changes

PR #1520, opened 2019-03-06
AUS: Updated neighbourhood records

stepps00:
Fixes https://github.com/whosonfirst-data/whosonfirst-data/issues/1521.

This PR includes:

- `mz:hierarchy_label` =0 flags for all neighbourhoods in Australia.
- If a neighbourhood has a Polygon or MultiPolygon geometry, the geometry was converted to a Point and the existing geometry was stored as an alt file.


nvkelso:
I'm going to merge this, no PIP is required (most these were points already, and for the polygons we don't have [microhoods in Australia](https://spelunker.whosonfirst.org/id/85632793/descendants/?exclude=nullisland&placetype=microhood)).

PR #1526, opened 2019-03-12
N/S Korea: Update DMZ geometries

### LINKED ISSUE #1510: Update geometry of the Korean Demilitarized Zone
https://spelunker.whosonfirst.org/id/1159339505/

screen shot 2019-03-04 at 11 43 02 am

The current geometry is generalized; create a new geometry and move the existing geometry to an alt file.

**See**: https://en.wikipedia.org/wiki/Korean_Demilitarized_Zone

> The DMZ is 250 kilometres (160 miles) long, and about 4 kilometres (2.5 miles) wide.

stepps00:
Fixes #1510 

This PR includes:
- Updated geometries of the two DMZ records in North Korea and South Korea
- Moved the existing geometries to alt files
- PIPs the records to update any hierarchies

The new geometries are not perfect, but they represent 2km buffer zones from the country boundaries (with adjustments on the west end). The DMZ is roughly 4km wide, but differs in some areas.

To do:
- [x] Validate changes


nvkelso:
Generally a lot better!

I wonder if the western portions are necessary at all as the water provides separation?

stepps00:
I'm going to merge as is, but we can always update the geometry/store a new "clipped" version in the future.

stepps00:
Geometries and files validate, merging.

PR #1531, opened 2019-03-18
Fix Greater Poland Voivodeship region eng name

mleginus:
While some regions in Poland do have English name correctly translated like (Województwo małopolskie -> Lesser Poland Voivodeship / https://spelunker.whosonfirst.org/id/85687291/). This is not the case for Województwo wielkopolskie - > Greater Poland Voivodeship - https://spelunker.whosonfirst.org/id/85687323/.


data/856/873/23/85687323.geojson
```diff
@@ -58,7 +58,7 @@
         "Woiwodschap Groot-Polen"
     ],
     "name:eng_x_preferred":[
-        "Wielkopolskie"
+        "Greater Poland Voivodeship"
```
nvkelso: Normally we'd say `Greater Poland`(not including the placetype in the default preferred localized name) and move the full `Greater Poland Voivodeship` to something like `label:eng_x_preferred` or `label:eng_x_preferred_longname`?

@stepps00 what's your recommendation?

nvkelso:
@mleginus thanks for your PR and issue! We'll have a look :)

data/856/873/23/85687323.geojson
stepps00: This full name would go in `label:eng_x_preferred_longname`, we'd also save the existing preferred name in the `name:eng_x_variant` property and update the `wof:name` accordingly.

nvkelso:
Closed in favor of https://github.com/whosonfirst-data/whosonfirst-data/pull/1533.

PR #1537, opened 2019-03-20
USA: Update hierarchy_label neighbourhood flag

stepps00:
Gowanus Heights is already flagged as `mz:is_funky` = 1. 

This PR updates the `mz:hierarchy_label` flag to = 0.


nvkelso:
This is a useful "tell" for the data, I'd rather not make this change.

PR #1552, opened 2019-03-26
rename county Eßlingen to Esslingen

tadjik1:
according to [wiki](https://de.wikipedia.org/wiki/Landkreis_Esslingen) the official
name of this county is `Esslingen` in many european languages


data/102/063/329/102063329.geojson
```diff
@@ -17,19 +17,19 @@
     "mz:hierarchy_label":1,
     "mz:is_current":1,
     "name:deu_x_preferred":[
-        "E\u00dflingen"
+        "Esslingen"
```
nvkelso: Since `"E\u00dflingen"` was a legit name in German before the 1960s, and to respect our versioning practice where both names should be present in the record else a new ID is needed, please also store `"E\u00dflingen"` as a `"name:deu_x_alternate":`.

tadjik1:
hi @nvkelso , thanks a lot for your feedback. I've added alternate name for this entity.

nvkelso:
@stepps00 Can you look exportifying this and then we can merge?

tadjik1:
@nvkelso I will close this PR then. thanks!

stepps00:
Thanks @tadjik1 - we merged the fix today via #1557.

PR #1569, opened 2019-04-01
Fix typo in Bucharest city name

mihneadb:
It is Bucharest, not Bucharesti. https://en.wikipedia.org/wiki/Bucharest


stepps00:
Thanks @mihneadb!

stepps00:
Added these changes via https://github.com/whosonfirst-data/whosonfirst-data/pull/1572, closing.

PR #1595, opened 2019-04-24
Array nesting

missinglink:
Fix nested array


missinglink:
Is this something I can do from a patch inside GitHub or do I need a local checkout?

nvkelso:
Exportify is General WOF tool, you could run it on local copy of file and then paste into github.com and we’d be fine. Checkout is “better”, but takes a while these days. Stephen is OOO this week.

thisisaaronland:
Tangentially related, but still not ready to replace the `wof-exportify` Python tool (and it's small universe of dependencies is this:

https://github.com/whosonfirst/go-whosonfirst-export

stepps00:
Replaced with #1601. Going forward, the plan is to commit to per-country repos.

